PR #3, opened 2021-12-14
di: define vertical directions and axis helper

whereswaldon:
I keep finding places where I need to choose something based on the axis of text, and the ergonomics of doing that right now are pretty bad. This commit introduces the constants for expressing vertical text and an easy way to look up the axis of text.

I'm open to other ways of doing this if you see alternatives.

### DIFF
--- a/di/direction.go
+++ b/di/direction.go
@@ -1,10 +1,33 @@
 package di
 
-type Direction int
+// Direction indicates the layout direction of a piece of text.
+type Direction uint8
 
 const (
+	// DirectionLTR is for Left-to-Right text.
 	DirectionLTR Direction = iota
+	// DirectionRTL is for Right-to-Left text.
 	DirectionRTL
+	// DirectionTTB is for Top-to-Bottom text.
+	DirectionTTB
+	// DirectionBTT is for Bottom-to-Top text.
+	DirectionBTT
+)
 
-// Could add Top-to-Bottom (TTB) and Bottom-to-Top (BTT) in future.
+// Axis returns the layout axis for d.
+func (d Direction) Axis() Axis {
+	switch d {
+	case DirectionBTT, DirectionTTB:
+		return Vertical
+	default:
+		return Horizontal
+	}
+}
+
+// Axis indicates the axis of layout for a piece of text.
+type Axis uint8
+
+const (
+	Horizontal Axis = iota
+	Vertical
 )
--- a/di/direction.go
+++ b/di/direction.go
@@ -14,8 +14,21 @@ const (
 	DirectionBTT
 )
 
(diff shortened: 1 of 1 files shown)


stuartmscott (approved):

benoitkugler:
I also think it is a very useful feature. 

Do we need to introduce a new type though ? Wouldn't be a `IsHorizontal() bool` method sufficient ? My point is that there is only two choices, so that a `bool` seems a better fit. 
The use cases I've encountered so far are often of the form 
```
if IsHorizontal() {
} else {
}
```
which seems a bit more natural than 
```
if Axis() == Horizontal {
} else if Axis() == Vertical {
} else {
// actually never happens
}
```

This is a minor detail though, I'm happy with the current PR !

whereswaldon:
I see your point @benoitkugler, but I also prefer having an explicit type for the axis so that a function signature can ask for an axis rather than a direction (often only the axis matters).

I think I've found something that should make us both happy though:

```go
type Axis bool

const (
	Horizontal Axis = true
	Vertical   Axis = false
)

func main() {
	ax := Vertical
	if ax {
		fmt.Println("horizontal")
	} else {
		fmt.Println("vertical")
	}
}
```
https://go.dev/play/p/nV-dq0POwee

The most important thing is definitely not forcing users to add a third case to all of their conditional logic to handle something that cannot actually happen, and this solution achieves that.

benoitkugler:
> I think I've found something that should make us both happy though:

> type Axis bool

Indeed this seems perfect to me !

whereswaldon:
I've pushed a commit implementing the new representation. Please note that I flipped the definitions from the sample above. Horizontal is represented by false, and the method to retrieve the axis is called `IsVertical`. This aligns the zero value for an `Axis` with the most common case (horizontal text).

di/direction.go
```diff
@@ -1,10 +1,46 @@
...
+//     } else {
+//         // Must be Horizontal.
+//     }
+//
+//     switch d.IsVertical() {
+//         case Vertical:
+//         case Horzontal:
+//     }
+//
+func (d Direction) IsVertical() Axis {
```
andydotxyz, line 31: This all looks great about from the `Is...` returning not `bool`, it just doesn't seem to read correctly.
I agree with the idea of `Horizontal` and `Vertical` for other cases - but isn't using that type in result of a boolean operation a little strange?

whereswaldon: I guess it is odd. I've created a separate method that returns the axis, and changed the signature to return boolean.

andydotxyz (approved):
Thanks, I think that is clearer.
Feel free to disagree with me btw :)

PR #4, opened 2021-12-14
Font fallback

benoitkugler:
This is a try at font fallback. 

More tests could be added, maybe with real world examples (say two fonts and a rune sequence where fallback is expected).
Feel free to comment and link your test cases if you want me to add them.

Close https://github.com/go-text/shaping/issues/8.


README.md
```diff
@@ -0,0 +1,3 @@
+# typesetting
+
+This typesetting library is shared by multiple Go UI toolkits including Fyne, and GIO.
```
whereswaldon, line 3:
```suggestion
This library provides typesetting capabilities in pure Go. It is appropriate for use in GUI applications, and is shared by multiple Go UI toolkits including [Fyne](https://fyne.io) and [Gio](https://gioui.org).
```

shaping/input.go
```diff
@@ -33,3 +35,85 @@ type Input struct {
...
+// SplitByFontGlyphs split the runes from 'input' to several items, sharing the same
+// characteristics as 'input', expected for the `Face` which is set to
+// the first font among 'availableFonts' providing support for all the runes
+// in the item.
+// Runes supported by no fonts are mapped to the first element of 'availableFonts', which
+// must not be empty.
+// The 'Face' field of 'input' is ignored: only 'availableFaces' are consulted.
+// Rune coverage is obtained by calling the NominalGlyph() method of each font.
+func SplitByFontGlyphs(input Input, availableFaces []font.Face) []Input {
+	var splittedInputs []Input
```
whereswaldon, line 48:
```suggestion
	var splitInputs []Input
```

shaping/input.go
```diff
@@ -33,3 +35,85 @@ type Input struct {
...
+
+// SplitByFontGlyphs split the runes from 'input' to several items, sharing the same
+// characteristics as 'input', expected for the `Face` which is set to
+// the first font among 'availableFonts' providing support for all the runes
+// in the item.
+// Runes supported by no fonts are mapped to the first element of 'availableFonts', which
+// must not be empty.
+// The 'Face' field of 'input' is ignored: only 'availableFaces' are consulted.
+// Rune coverage is obtained by calling the NominalGlyph() method of each font.
+func SplitByFontGlyphs(input Input, availableFaces []font.Face) []Input {
```
whereswaldon, line 47: I could see an argument that we should create some kind of intermediate type like:

```go
type Run struct {
    Text []rune
    Start, End int
}
```

We could easily embed this inside of the `Input` type, but use it as the input to this function.

Then again, we probably want to ensure that the other fields of the returned `[]Input`s all match (size, script, language, direction), so it probably isn't a net win.

I don't really think we should change this, but I'll leave this comment in case it sparks thoughts for others.

shaping/input.go
```diff
@@ -33,3 +35,85 @@ type Input struct {
...
+//
+// We don't want to change fonts just for variation selectors.
+//
+// Finally, don't change fonts for line or paragraph separators.
+//
+// The rationale is taken from pango : see bugs
+// https://bugzilla.gnome.org/show_bug.cgi?id=355987
+// https://bugzilla.gnome.org/show_bug.cgi?id=701652
+// https://bugzilla.gnome.org/show_bug.cgi?id=781123
+// for more details.
```
whereswaldon, line 109: I'm _sooooo_ glad that you're the one who wrote this @benoitkugler! Having your experise about pango and various problems within it is proving quite invaluable! Thank you for including links to the various edge case bugs in pango.

shaping/input_test.go
```diff
@@ -0,0 +1,162 @@
+package shaping
```
whereswaldon, line 1: For the tests, what would you think about importing something like `gioui.org/font/gofont` and testing support of an actual font against some multi-lingual text? There's also `eliasnaur.com/roboto`. These test cases are excellent unit tests, but I think we could stand to have one or two tests that use a real font. It doesn't need to be the ones I mentioned; they would just be easy because they're importable packages.

whereswaldon (commented):
Overall, this is an excellent and extremely readable change. I'm amazed at how quickly you put this together!

I have a few tiny comments inline:
- variable name suggestion
- README text tweaks
- some thoughts on the approach
- ideas for broader testing

None of my comments are big though. I'd be pretty happy even if you didn't change anything. :D

shaping/input_test.go
benoitkugler: I also think we should add tests with a real font. 
Should we favor (open source) font files (to be loaded with textlayout/fonts) to avoid adding a dependency ? 
Would it be possible for you to provide some text input ? Otherwise I can try and build some multi-lingual text as you suggested.

whereswaldon (approved):
Great work!

andydotxyz (approved):
Cool, great addition thanks.

shaping/input.go
```diff
@@ -33,3 +35,85 @@ type Input struct {
...
+func ignoreFaceChange(r rune) bool {
+	return unicode.Is(unicode.Cc, r) || // control
+		unicode.Is(unicode.Cf, r) || // format
+		unicode.Is(unicode.Cs, r) || // surrogate
+		unicode.Is(unicode.Zl, r) || // line separator
+		unicode.Is(unicode.Zp, r) || // paragraph separator
+		(unicode.Is(unicode.Zs, r) && r != '\u1680') || // space separator != OGHAM SPACE MARK
+		('\U000e0100' <= r && r <= '\U000e01ef') ||
+		('\ufe00' <= r && r <= '\ufe0f') // variation selectors
+}
```
khaledhosny, line 119: I think a better source for these code points would be https://github.com/harfbuzz/harfbuzz/blob/067f90a82087a5feb1cf76625c54c60a10946fe1/src/hb-unicode.hh#L131-L200. The code from Pango will not enable fallback for U+06DD (a long outstanding bug in Pango) because Unicode unhelpfully categorize it as format character, even though it always have a graphical representation.

benoitkugler: Thanks for your input !

To be certain, you suggest to ignore the code points in `is_default_ignorable`, as well as in the categories : 

- Separator, {Space, Line, Paragraph}
- Other, {Control, Surrogate}

(the correct code points in category Other, Format being handled by `is_default_ignorable`) 

Did I get this right ?

khaledhosny: Yes, sounds right.

PR #5, opened 2021-12-15
ci: add basic compilation and test running to CI

whereswaldon:
I figured it would be nice to have a little CI. This is just a really basic config. I based it off of [this](https://github.com/mvdan/github-actions-golang), but I removed platforms and go versions that are probably overkill as testing targets. Right now, the fact that it's a matrix build does us no good, but leaving that structure would make it easy to insert new testing go versions and OSes if we wanted to. I don't mind pulling it out though.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,18 @@
+on: [push, pull_request]
+name: Test
+jobs:
+  test:
+    strategy:
+      matrix:
+        go-version: [1.17.x]
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ matrix.go-version }}
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Test
+      run: go test ./...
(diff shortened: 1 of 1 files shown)


stuartmscott (approved):
Looks good!

Any value in adding fmt, vet, or lint as well?

whereswaldon:
All of the above are good, but I just wanted something quick, simple, and uncontentious first. Especially with linters, there's lots of room to disagree about what's worth enforcing.

.github/workflows/test.yml
```diff
@@ -0,0 +1,18 @@
+on: [push, pull_request]
+name: Test
+jobs:
+  test:
+    strategy:
+      matrix:
+        go-version: [1.17.x]
```
andydotxyz, line 7: Can we test with more than just the latest Go?
I don't know what version people are comfortable with. Fyne supports back to 1.14 currently so we can build for iOS 32bit devices.
I guess we should at least test against 1.16 or match what the Go team support?

whereswaldon: Go supports the last two minor releases at any given time (so 1.16. and 1.17 right now). I'm comfortable matching their support level, but not going all the way back to 1.14.

whereswaldon: I've added 1.16 to the matrix.

andydotxyz (approved):
OK thanks.

PR #6, opened 2021-12-15
Remove public external references

andydotxyz:
A minor change in concept - remove the external dependency references from our public API.

Following this developers can purely use the `go-text/typesetting` without knowing about `harfbuzz` or `fonts` packages from external repo.

Discussions may be required about packages or naming, but as discussed elsewhere the API is not final/stable yet.


whereswaldon (changes requested):
Looking solid! I have one concern, but it's about placement, not approach. Nice work.

andydotxyz:
> Looking solid! I have one concern, but it's about placement, not approach. Nice work.

OK no problems. But I can't see the are you'd like me to change, sorry.
Am I missing something in the review data?

font/language.go
```diff
@@ -0,0 +1,6 @@
+package font
+
+import "github.com/benoitkugler/textlayout/language"
+
+type Language = language.Language
+type Script = language.Script
```
whereswaldon, line 6: My only concern is that these are not inherently a font characteristic, but rather a characteristic of the input text. I'd prefer to move these into a separate `language` package in `typesetting` or put them in `shaping` (since that's where they're consumed from). As it stands, I think accessing them as `font.{Language,Script}` is misleading.

I totally typed this out before, but I think maybe I didn't finalize the comment and then I submitted my review and the draft vanished. Sorry for the confusing review with no context!

whereswaldon (changes requested):
I _hope_ my comment shows up this time. :D

font/language.go
andydotxyz: I see this now, thanks.
Same has happened to me before :)

font/font.go
```diff
@@ -0,0 +1,13 @@
+package font
+
+import (
+	"github.com/benoitkugler/textlayout/fonts"
+	"github.com/benoitkugler/textlayout/fonts/truetype"
+)
+
+type Font = *truetype.Font
```
benoitkugler, line 8: This naming suggest that we only support truetype/opentype font files.

I seems to me than some people have expressed interest for older font formats. We probably should discuss what formats we want to support, but in the meantime, I suggest renaming to `OpentypeFont` for instance.

font/face.go
```diff
@@ -1,13 +1,16 @@
...
+import (
+	"github.com/benoitkugler/textlayout/fonts"
+	"github.com/benoitkugler/textlayout/harfbuzz"
+)
 
 type GID = fonts.GID
+type GlyphMask = harfbuzz.GlyphMask
 
 type Face interface {
+	fonts.Face // make sure that we can use textlayout/fonts.Face as Face
```
benoitkugler, line 12: Minor comment : I was thinking to `textlayout/fonts.Face` as a draft interface, that we will polish when building go-text. 
(When the API is settled, it is not obvious that the methods in  `textlayout/fonts.Face` are all needed.)

Anyway, if we prefer to stick to `textlayout/fonts.Face`, we could simply use 
`type Face = textlayout/fonts.Face`

benoitkugler (changes requested):
Very nice work (I now better understand what you had in mind). It is indeed a much cleaner external API !
I have two comments : one major, one minor.

font/font.go
andydotxyz: There certainly is some confusion about Font vs Face which led to me taking this approach.
Going back to just `Face` for now to avoid the implication of only truetype support.

font/face.go
andydotxyz: I started with a `Face` alias - but then `NominalGlyph` was added, and I didn't think we wanted two different face interfaces in this package?

whereswaldon:
I'm happy once @benoitkugler is happy. :D

font/face.go
benoitkugler: Alright. Since the `NominalGlyph` is already in `textlayout/fonts.Face`, I then suggest to use `type Face = textlayout/fonts.Face`

benoitkugler:
LGTM ! Thank you for the changes.

benoitkugler (approved):

font/font.go
```diff
@@ -0,0 +1,12 @@
+package font
+
+import (
+	"github.com/benoitkugler/textlayout/fonts"
+	"github.com/benoitkugler/textlayout/fonts/truetype"
+)
+
+type Resource = fonts.Resource
+
+func ParseTTF(file Resource, loadAllTables bool) (Face, error){
```
benoitkugler, line 10: Actually, in the context of this package, we always want to load all the tables.
So I would remove `loadAllTables` from the `ParseTTF` arguments (and pass `true` to the  `truetype.Parse` call.

benoitkugler (changes requested):
Sorry, one last minute detail..

whereswaldon (approved):

benoitkugler (approved):

PR #13, opened 2021-12-17
Clean the recompute API

benoitkugler:
With #12 merged, we can simplify the Output.{RecomputeAdvance, RecalculateAll} signatures.
(This is safe because we never wan't to use a different direction than the once used to shape the input.)


whereswaldon (changes requested):
Could we simplify the switches to be Axis-based instead of Direction-based? I think it would read a little better.

benoitkugler:
Sure ! (I've used the if `IsVertical` form.)

whereswaldon (approved):

andydotxyz (approved):
Nice tidy

PR #15, opened 2021-12-20
di: add text progression types and method

### LINKED ISSUE #14: Proposal: Direction attribute for text progression
The four Directions defined in `di` have two attributes:

- Are they vertical or horizontal?
- Does the text "begin" at the origin and move away along the axis, or does it begin away from the origin and move towards it? (assuming the origin is in the upper left).

There are many times when it would be convenient to define an operation based on this second property, as it would allow LTR and TTB to share an implementation of an algorithm, and would likewise allow sharing between RTL and BTT.

I'd like to add something like the `Axis` method to `di.Direction` to return this attribute, but I honestly have no idea what to call this property.

Do you think this makes sense and is a good idea? Is there already a typographical name for this aspect of text direction?

whereswaldon:
This commit introduces a type that represents how a piece
of text is read relative to the origin as well as a
method to get the progression from any direction.

This information makes it easy to write algorithms that
operate on text of only one progression, but across any
axis.

Closes #14

### DIFF
--- a/di/direction.go
+++ b/di/direction.go
@@ -35,9 +35,9 @@ func (d Direction) Axis() Axis {
 func (d Direction) Progression() Progression {
 	switch d {
 	case DirectionTTB, DirectionLTR:
-		return FromOrigin
+		return FromUpperLeft
 	default:
-		return TowardOrigin
+		return TowardUpperLeft
 	}
 }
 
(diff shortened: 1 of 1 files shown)


andydotxyz:
I'm a little confused here, I am not sure if it relates to naming or purpose.
It looks like RTL is always defined as "toward origin" - but in a RTL document the origin is in a different place, so it will be "from origin".
So I guess the question is, does this need to take the context of the overall document, or does it need naming change where origin is more specifically top left?

benoitkugler:
> I'm a little confused here, I am not sure if it relates to naming or purpose. It looks like RTL is always defined as "toward origin" - but in a RTL document the origin is in a different place, so it will be "from origin". 

Well, following the context of #14,  the origin of _the (drawing) pen_ would change, but not the origin of the document. But maybe it is not clear enough.

whereswaldon:
@benoitkugler @andydotxyz Does this naming work for everyone?

benoitkugler:
> @benoitkugler @andydotxyz Does this naming work for everyone?

I'm happy with it, thanks !

andydotxyz:
Sorry I missed this, was swamped with other things leading to the holiday.
This is much clearer, though reading through it I wondered - should it be "Top" instead of "Upper"? That way it matches the TTB and LTR naming...

whereswaldon:
I think we have enough consensus on this to merge this version.

andydotxyz:
> I think we have enough consensus on this to merge this version.

Personally I would have waited for @benoitkugler, though I was going to approve the current state anyhow.

whereswaldon:
I figured that the difference between "Upper" and "Top" was very slim, and an approval for "Upper" probably translated. If @benoitkugler prefers something else, I'm happy to have another PR change it.

Issue #17, opened 2022-01-02
Add support for font selection


benoitkugler:
# Proposal 

This is a draft proposal to add support for font discovery and semi-automatic font selection to go-text. 

## The problem to solve

One of the task go-text is currently not able to perform is to choose a font to use for a text run, given a font description. This is required for instance to implement rendering of HTML files (which is my main use case for go-text).

Before outlining the proposed solution, let me precise what I mean by "font description" and "choose a font".

### Font description

A font description is a possibly incomplete list of attributes that describe the intention of a document author about the text font.

More precisely,  using the CSS terminology, a font description is composed of a font _family_ name (or a special keyword to describe commonly used families) and some style attributes, like _stretching_, _weight_ and _style_.

### Font resolution

When rendering a document, a layout engine has to resolve this description against the actual fonts it has at its disposal. This requires a list of the font files and for each of them, several properties such as
 - the properties listed in the font description
 - the rune coverage (a set of Unicode codepoints supported by the font)
 - the language coverage (usually deduced from the rune coverage)

## The issue

Using a font library, you can pretty easily query the information mentioned above. However, this would require to load every font files on your machine (many hundreds for standard systems). 

## One solution

The solution used on Linux systems, implemented in the fontconfig library, is to build once for all an index, that is a list of footprints or patterns (one for every font), which are lightweight summaries of the font content. Building this index is time consuming, but loading it from the disk is really fast, meaning that once it has been build, you don't have to worry about performance anymore. 
The second goal of the library is then to be able to resolve a font file from a font description. This is roughly speaking implemented by comparing the font description against every footprint (which again is very efficient since you don't have to load the font file in memory).

## The actual proposal for go-text

I propose to implement and add to go-text a simplified version of fontconfig, with two main parts :
 - the tooling needed to query and build the font index 
 - an API to perform the resolution of a font description against such an index. 

### A possible implementation sketch

As implementation guidelines, I would follow the fontconfig implementation for the first part, and the CSS specification (see font-kit) for the second.

As a starting point, we could use something like the following types : 

```
type FontDescription struct {
     Family []string // many families possible
     Style 
     Stretch
     Weight
 }

type Footprint struct {
       Source Source
       Family string
       Style
       Stretch
       Weight
       RuneCoverage 
       LanguageCoverage
}   
     
type Source struct {
    FileID string
    FontIndex int // to support font collections
    InstanceIndex int // to support variable fonts
    FontFormat // optional, to make font loading easier if we support multiple font formats
[...]

PR #18, opened 2022-01-07
[shaping] add a hook for a general face selection

benoitkugler:
One of my main motivation to discuss #17 is that I initially thought the shaping API would have to know about font descriptions to support the general capabilities I need. 

I now realize that it actually only requires a small addition to the shaping package. With this PR, the font index mechanism described in #17 may then entirely be implemented outside shaping (or even outside go-text).

Notice that I've written this PR to be entirely backward compatible. 
As a result, `SplitByFontGlyphs` and `SplitByFace` are somewhat redundant. We could remove `SplitByFontGlyphs` and expose the `fixedFontmap` type, as you wish.

### DIFF
--- a/shaping/input.go
+++ b/shaping/input.go
@@ -37,6 +37,29 @@ type Input struct {
 	Language language.Language
 }
 
+// Fontmap provides a general mechanism to select
+// a face to use when shaping text.
+type Fontmap interface {
+	// ResolveFace is called by `SplitByFace` for each input rune potentially
+	// triggering a face change.
+	// It must always return a valid (non nil) font.Face value.
+	ResolveFace(r rune) font.Face
+}
+
+var _ Fontmap = fixedFontmap(nil)
+
+type fixedFontmap []font.Face
+
+// ResolveFace panics if the slice is empty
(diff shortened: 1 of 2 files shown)


whereswaldon (approved):
Looks good to me. Nicely done!

andydotxyz:
The code looks fine - I was surprised that you can resolve the hint based lookup you needed with this simple API.
Isn't it also necessary to know the offset into `Input` that the `rune` was positioned at so you can look up your hint data?

benoitkugler:
> The code looks fine - I was surprised that you can resolve the hint based lookup you needed with this simple API. Isn't it also necessary to know the offset into `Input` that the `rune` was positioned at so you can look up your hint data?

In the use case I have in mind, I would start by splitting the input text according to the font hint. Then, I would fetch the faces for one hint, and call the SplitByFace with the corresponding font map (then repeating this for each font hint run).
This means that I would only call go-text with text associated with a single font hint, but it seems to me that it is not a problem.

andydotxyz:
OK, thanks for the explanation - this makes sense.

andydotxyz (approved):

PR #19, opened 2022-03-18
shaping: implement line wrapping

whereswaldon:
As promised, here's the line wrapping algorithm that currently powers Gio's usage of our typesetter. Please note that I'm not married to the names of symbols in this API, and that nothing is exported right now because I'm not entirely certain what we want to expose.

One detail that I want to call attention to is the `output` type. It wraps our `Output` with information about which runes of input correspond to the output. This information was critical for me to be able to correctly map sequences of glyphs to runes in Gio's text editor. I'm happy to change types around, but I want to preserve that somehow. Maybe the right way is simply to add the fields to the normal `Output` type.

I've tried to frame line wrapping as a simple yes/no decision problem in this implementation. The function `shouldKeepSegmentOnLine` is that decision. In English, the decision is "should I keep the segment from the current line segment to the given line-break candidate on this line? (given some space constraints)". The decision itself can be made much smarter (it doesn't even break on non-word boundaries right now), but it seems like it might be a good starting point.

The `lineWrap` function is the real entry point into all of the logic. It's not currently exported because I'm not sure it's the interface we want to export.

I should also note that this simple strategy does not handle the following:
- does not currently enforce mandatory UAX#14 breaks, though it has the information necessary to do so.
- does not check for harfbuzz unsafe to break markers on the glyph clusters adjacent to break candidates

It does, however:
- reject line break candidates that would break a harfbuzz cluster
- properly break RTL and LTR text

~As for the many lines of new `go.sum`, I'm [hoping we can do something about that](https://github.com/npillmayer/uax/issues/10).~ I've pushed an update that switches to a zero-dep implementation of UAX.

I welcome your thoughts as you have time to look at it!


benoitkugler:
I'm starting to dig deeper in this PR; thanks for this large work !
One of my first impression is that we should handle the case of wrapping a multi-`Output`. I've two use cases in mind : 

- A line of text with mixed (say) Latin and Arabic will be segmented in two `Input`s, yielding two `Output`s, but line wrapping needs both `Output`s, right ? 
Unless there is a way to call `lineWrap` on each `Output`, adjusting `maxWidth`, but it feels a bit clunky.
- Similarly, a RTL text embedded in a LTR will yield two (or more) `Output`s

What do you think ? We perhaps should use of a new type representing this new level of abstraction : `Output` is a single run in a line, it is linked to Harfbuzz requirements, and the new type, say `LineOutput`, would basically be a slice of `Output`s, representing a line of text (before line wrapping). Then `lineWrap` acts on `LineOutput`s rather than single `Output`s.
I've not thought of the complication in `lineWrap` though...

whereswaldon:
> What do you think ? We perhaps should use of a new type representing this new level of abstraction : `Output` is a single run in a line, it is linked to Harfbuzz requirements, and the new type, say `LineOutput`, would basically be a slice of `Output`s, representing a line of text (before line wrapping). Then `lineWrap` acts on `LineOutput`s rather than single `Output`s. I've not thought of the complication in `lineWrap` though...

Hi @benoitkugler , I totally agree that the desired final behavior is having the line wrapper accept `[]shaping.Output` and return `[]shaping.Output`. However, that's yet another hike in complexity. The approach that I've taken is to tackle the simpler (but still hard) case of wrapping a uniform run of text correctly (based on its direction), and then we can look to upgrade to wrapping multiple outputs. Would you be okay reviewing this with the lens that it is only attempting to implement that simpler case?

benoitkugler:
> Would you be okay reviewing this with the lens that it is only attempting to implement that simpler case?

Sure ! Not exporting the current symbols makes a lot of sense then. The current tests will also be very helpful when trying to upgrade to multi-output (because such a change should not change the results for one output.)

shaping/wrapping.go
```diff
@@ -0,0 +1,243 @@
...
+		end := b.breakAtRune
+	innerLoop:
+		for k := i + 1; k < len(breaks); k++ {
+			bb := breaks[k]
+			candidate, ok := shouldKeepSegmentOnLine(out, runeToGlyph, start, bb, maxWidth, good.Advance.Ceil(), maxWidth)
+			if ok {
+				// Use this new, longer segment.
+				good = candidate
+				end = bb.breakAtRune
+				i++
```
benoitkugler, line 219: Could you add a comment explaining why we go through ?

shaping/wrapping.go
```diff
@@ -0,0 +1,243 @@
...
+			} else {
+				break innerLoop
```
benoitkugler, line 221: Could you add a comment explaining why we break ?

shaping/wrapping.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	start := 0
+	runesProcessed := 0
+	for i := 0; i < len(breaks); i++ {
+		b := breaks[i]
+		// Always keep the first segment on a line.
+		good, _ := shouldKeepSegmentOnLine(out, runeToGlyph, start, b, maxWidth, 0, maxWidth)
+		end := b.breakAtRune
+	innerLoop:
```
benoitkugler, line 211: I think the label is not needed.

shaping/wrapping.go
```diff
@@ -0,0 +1,243 @@
...
+// inclusiveGlyphRange returns the inclusive range of runes and glyphs matching
+// the provided start and breakAfter rune positions.
+// runeToGlyph must be a valid mapping from the rune representation to the
+// glyph reprsentation produced by mapRunesToClusterIndices.
+// numGlyphs is the number of glyphs in the output representing the runes
+// under consideration.
+func inclusiveGlyphRange(start, breakAfter int, runeToGlyph []int, numGlyphs int) (glyphStart, glyphEnd int) {
+	rtl := runeToGlyph[len(runeToGlyph)-1] < runeToGlyph[0]
+	runeStart := start
+	runeEnd := breakAfter
```
benoitkugler, line 65: Are the new variables really needed ?

benoitkugler (changes requested):
I've finally took some time to review this PR, thank you again for this nice implementation.

I have to confess I have not grasped all the subtleties of the algorithm, but I don't see anything wrong.

I have proposed in #21 some surface refactoring to help understanding the logic (and also provides a solution to my review comments). Of course, feel free to ignore them if you have a strong opinion about it.

shaping/wrapping.go
whereswaldon: done!

shaping/wrapping.go
whereswaldon: done!

shaping/wrapping.go
whereswaldon: Agreed. The logic went through many iterations, and I suspect this is an artifact left behind.

shaping/wrapping.go
whereswaldon: Nope, these are another artifact of an older version of the code.

whereswaldon:
@benoitkugler Thanks a bunch for the review. I've adopted your requested changes, and I've also pushed some work I had locally that (I think) further simplifies things. The line break candidates are now generated iteratively as needed instead of up-front. This should allocate less memory and be a little easier to think about.

I also fixed a bug in the tests (there's a separate PR with just that fix available in #22 ) and updated the Go versions we use in CI to ones that support math.MaxInt (1.17+).

whereswaldon:
I think I'd really like to unify `output` and `Output`. The rune range seems (to me) to be a totally reasonable thing to include. Thoughts?

benoitkugler:
> I think I'd really like to unify `output` and `Output`. The rune range seems (to me) to be a totally reasonable thing to include. Thoughts?

Yes, good point, let's do it.

benoitkugler (approved):

whereswaldon:
> > I think I'd really like to unify `output` and `Output`. The rune range seems (to me) to be a totally reasonable thing to include. Thoughts?
> 
> Yes, good point, let's do it.

Done!

While we're at it... Maybe we should export the `shapedParagraph` type (call it just `Paragraph` or something?)? It seems like an okay starting point for a public API. I think it will definitely need to change, but it's usable as-is.

You just do:

```go
// assume we have input
var input shaping.Input
shapedOutput, err := shaping.Shape(input)
if err != nil {
    // handle
}
lines := shaping.NewParagraph(shapedOutput).LineWrap(width)
```

benoitkugler:
Exposing a Paragraph type seems okay to me.
I think it could later be the place to put options influencing line breaking like hyphenation.
As you said, it is definitely a good starting point.

By the way, have you already start to think about the multi input/output case ? Or do we stay on the single input case as you proposed earlier?

whereswaldon:
> Exposing a Paragraph type seems okay to me. I think it could later be the place to put options influencing line breaking like hyphenation. As you said, it is definitely a good starting point.

On second thought, I think we should call it a `Wrapper` or perhaps `LineWrapper` and surface a `WrapParagraph` method. I think those names match slightly better with what we already have. Later we could add a `WrapLine` to support the "I only want N lines" usecase that pango can't handle.

> By the way, have you already started to think about the multi input/output case ? Or do we remain focused on the single input case as you proposed earlier?

I've thought about it quite a bit, and I think I have a good idea how to go about it. However, it's a pretty involved change to a bit of code that took me nearly a month to get right, so it's not trivial. That's why [Gio is raising funds](https://opencollective.com/gioui/projects/font-fallback) to support the effort. The biggest/hardest change to support Gio gaining font fallback is updating the line wrapper to support multiple heterogeneous `shaping.Output`s as input.

whereswaldon:
I've pushed a minimal exported API like what we discussed, but I don't have strong opinions about the symbol names or structure.

benoitkugler:
@andydotxyz I share your concern about the dependency on `uax`.
I'm working on a drop in replacement, taking inspiration from its nice API but also from what I've learned studying Pango stack. 
I need to work on some edges cases, but I'm hoping to propose a PR soon (with more context of course)

benoitkugler:
@whereswaldon  indeed, the names are better, thanks !

whereswaldon:
@benoitkugler can you elaborate on your concerns with the UAX repo? It's license compatible, (now) zero new transitive dependencies, maintained (albeit by different people), very thoroughly tested, and optimized. We currently need only UAX#14 (line breaking), but we will need UAX#29 (grapheme cluster breaking) in order to properly break within words when we add that feature. I think the existing bidi segmentation algorithm in there may also be a good starting point for us when we try to support that, though I've been less focused on that case.

I'd rather see us improve using it as a starting point than redo all of the work. We can even move that repo into go-text. That being said, I don't yet know where your concerns are coming from. Perhaps you will change my mind. :D

As for the API and the lessons you learned from pango, I'd like to hear more. There are actually things I'd love to change about the UAX API.

go.mod
```diff
@@ -4,6 +4,6 @@ go 1.15
 
 require (
 	github.com/benoitkugler/textlayout v0.0.5
-	golang.org/x/image v0.0.0-20210607152325-775e3b0c77b9
-	golang.org/x/text v0.3.7 // indirect
+	github.com/gioui/uax v0.2.1-0.20220819135011-cda973fac06d
```
andydotxyz, line 7: It makes me nervous that a text project aiming to be toolkit agnostic would import repos from one of the toolkit accounts.
If it needs to be forked for better maintenance would this (`go-text`) not be the right place to do it?

whereswaldon: Gio forked and improved the code because Gio actively depends on it in our main branch. Since these changes were in limbo at the time, it wasn't clear to me that I should put it in go-text. That being said, I've already offered to move it to this org in the top level repo comments. There's nothing territorial here. It just made sense at the time to put it in Gio's org.

andydotxyz: Absolutely - I did not mean to imply it was territorial, just that it should be re-considered for the purpose of this PR.

whereswaldon: If we decide that the content of the dependency is okay, I'll move it into go-text before we merge this.

andydotxyz: I'd be interested to hear more from @benoitkugler - is their replacement a better alternative?

benoitkugler: I had a lot of free time last week, and I was able to work on the replacement implementation I was referring to.

To give more context and answer @whereswaldon question, I think we should use this over `github.com/gioui/uax` for two reasons : 

- The code in `uax` is very high level : the implementation strategy is very refined, but it makes the code quite complicated to understand. I'm not referring to the Unicode rules, some of them being inherently intricate, but rather to the general logic flow : you need to understand several concepts and structures (such as `NfaStateFn`, `Recognizer`, `UnicodePublisher`, `RuneSubscriber`, ...) to be able to understand the code. 
- The abstraction has a performance cost : even with the nice work on the `gioui` fork, both speed and memory allocations are significantly higher than a "simpler", more linear, approach

First precision : the performance argument is probably not very strong for now, but still I think that a low level, shared, text framework should aim at good performances.
Second precision : I actually didn't understand the logic in `uax`, which makes my first point a bit biased.. but I suspect I won't be the only one..

I have pushed two branches : 

- [segmenter](https://github.com/go-text/typesetting/tree/segmenter/segmenter) is a playground with both implementations available, so that you can easily run the benchmarks in the `segmenter` package, and check that my performance point is accurate
- [line-wrapping-segmenter](https://github.com/go-text/typesetting/tree/line-wrapping-segmenter/segmenter) uses  this new package as a replacement for `uax` with minimal changes (and tests passing).

The Unicode conformance tests for [Graphemes](https://www.unicode.org/Public/15.0.0/ucd/auxiliary/GraphemeBreakTest.txt) and [Line Breaks](https://www.unicode.org/Public/15.0.0/ucd/auxiliary/LineBreakTest.txt) are passing.

Regarding the license, I have borrowed the iterator pattern from `uax` and the concept of the break attributes from the Pango source code, but I've written the go code from the ground up, and thus the license should not be an issue here. 

I propose this package as an alternative, but I will let you decide what is best. 

(If using this new `segmenter` package sounds good to you, we should probably start by reviewing its code before  merging this PR.)

whereswaldon: I'm looking at your implementation now @benoitkugler and it's very impressive. I think there's little doubt that we should switch to it. I do think I'll have some questions/suggestions about the API though. I just need some more time to look and experiment.

benoitkugler: > I do think I'll have some questions/suggestions about the API though.

Sure thing, I'll be happy to include your suggestions !

whereswaldon:
This is obsoleted by #25

PR #21, opened 2022-09-15
minor refactors to line breaking logic

benoitkugler:
Some suggestions to hopefully make the logic a bit clearer.
No fundamental changes.
Names are open to discussion :)


whereswaldon:
Great work! I've incorporated these changes into #19, so I'm going to close this.

andydotxyz:
This PR added a new external dependency on a repository that has just 6 stars. Just wanted to mention that and confirm that everyone is ok introducing this dep?

It is also a BSD license which, as I understand it, means that an Unlicense toolkit may have to be clear that users are taking on a license more restrictive than the main project?

whereswaldon:
> This PR added a new external dependency on a repository that has just 6 stars. Just wanted to mention that and confirm that everyone is ok introducing this dep?

This PR was just a way for @benoitkugler to suggest changes to #19. I added the dependency there, so it's probably best to discuss it there.

> It is also a BSD license which, as I understand it, means that an Unlicense toolkit may have to be clear that users are taking on a license more restrictive than the main project?

That repository is a thoroughly tested implementation of many of the Unicode text segmentation algorithms, which is pretty hard to find. It is also dual-licensed BSD/UNLICENSE, just like our repo.

The exact repo we're depending on has changed since this PR, so I'd recommend reviewing the dependency changes on #19.

andydotxyz:
Apologies, I missed that there were two licenses in the LICENSE file.
But I did notice that readme states "This is very much work in progress, not intended for production use. Please be patient."

PR #22, opened 2022-09-15
shaping: fix invalid use of errors.As

whereswaldon:
Go 1.19 refuses to run the tests with this mistake in place.

### DIFF
--- a/shaping/output_test.go
+++ b/shaping/output_test.go
@@ -91,7 +91,7 @@ func TestRecalculate(t *testing.T) {
 		Direction di.Direction
 		Input     []shaping.Glyph
 		Output    shaping.Output
-		Error     error
+		Error     *shaping.UnimplementedDirectionError
 	}
 	for _, tc := range []testcase{
 		{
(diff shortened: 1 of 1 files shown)


benoitkugler:
LGTM

shaping/output_test.go
```diff
@@ -91,7 +91,7 @@ func TestRecalculate(t *testing.T) {
 		Direction di.Direction
 		Input     []shaping.Glyph
 		Output    shaping.Output
-		Error     error
```
andydotxyz, line 94: Was this line really required too? seems kinda limiting

whereswaldon: I couldn't see a simple change that didn't require this. What we want to achieve is checking whether the concrete type of a returned error matches the expected concrete type. That's actually kinda a pain. There's likely a way to do it with reflection, but there's no simple mechanism available. I figured that since we only need to check for one concrete type right now, we can do this. If we introduce other concrete types into the API, we can then bite the complexity bullet of implementing the reflect-based type matching.

PR #30, opened 2022-11-07
shaping: fix and test cluster splitting bug

whereswaldon:
I discovered that our logic to prevent splitting harfbuzz clusters during wrapping only worked if the run in consideration was at the beginning of the shaped text. A Gio fuzz test for rune accounting was failing as a result. I've fixed the logic and added a new test case to exercise the non-zero-offset cluster splitting case.


benoitkugler (approved):

andydotxyz (approved):
Good catch

PR #31, opened 2022-11-11
shaping: provide size in output

whereswaldon:
This commit copies the shaping.Input's size to the corresponding shaping.Output. This makes shaping text with variable-sized runs much easier, as you don't need to do manual tracking of the run sizes.

### DIFF
--- a/shaping/output.go
+++ b/shaping/output.go
@@ -86,6 +86,8 @@ func (b Bounds) LineHeight() fixed.Int26_6 {
 type Output struct {
 	// Advance is the distance the Dot has advanced.
 	Advance fixed.Int26_6
+	// Size is copied from the shaping.Input.Size that produced this Output.
+	Size fixed.Int26_6
 	// Glyphs are the shaped output text.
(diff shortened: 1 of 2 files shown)


benoitkugler (approved):

shaping/output.go
```diff
@@ -86,6 +86,9 @@ func (b Bounds) LineHeight() fixed.Int26_6 {
 type Output struct {
 	// Advance is the distance the Dot has advanced.
 	Advance fixed.Int26_6
+	// Size is the scaling factor applied to this text during shaping, copied from
```
andydotxyz, line 89: Is it just me that is confused by this documentation?
I went to the Input struct to see what it means and I do understand it now, but to state that size is a scaling factor alone seems potentially confusing.

whereswaldon: I've updated it to simply state that the field is copied from `shaping.Input.Size`. That way we only need to document what it _means_ in one place.

andydotxyz (approved):

PR #34, opened 2022-12-09
[deps] update to last textlayout version

benoitkugler:
@Jacalz has been working on binary size and memory optimisation in textlayout, available in v0.3.0


whereswaldon (approved):
Thank you!

andydotxyz (approved):
Excited that we can trim a few things down :)

Issue #28, opened 2022-10-12
How to implement line wrapping parameters?


whereswaldon:
So I think we're clearly going to want to be able to configure the line wrapper's behavior. Things like:

- enable/disable breaking lines within a word, and various heuristics for when to do so
- hyphenation
- maybe justification? This *seems* like it can be implemented as a separate step for most scripts, just spreading extra space across whitespace glyphs. For scripts that don't use whitespace, I frankly have no idea how to do this. I think we can deal with that later though.
- allowing re-shaping segments to enable breaking at previously unbreakable places?

I'm not asking for an inventory of these parameters or anything, but rather for thoughts on how they should be provided. My inclination is to add something like:

```go
type WrapConfig struct {
    BreakWordPolicy // Type encapsulating the heuristics for how to break within a word
    Hyphenate bool
    Hyphen Output // The actual hyphen glyph to use, the result of a previous shape
    // maybe other fields later
}
```

This could then become a new parameter to `LineWrapper.Prepare` and `LineWrapper.WrapParagraph`. That would create a constraint that the configuration must be constant for the duration of a paragraph, but I think that's probably desirable. Wrapping individual lines with different choices seems... strange?

The other obvious API choice would be adding this as a field on the LineWrapper, but I think that makes reusing a single LineWrapper instance for many different pieces of content unnecessarily hard.

I think the most urgently needed option here is enabling the ability to break within a word. Right now Gio simply can't do it (because we don't surface it), so strings in extremely narrow spaces are just visually truncated by-word. You get to see the beginning of each word in the string. It's pretty sub-optimal. :D

benoitkugler:
Another possible parameter would be related to ellipsis (that is, add ... and truncate instead of wrapping).

Besides, I'm currently thinking about two possible sources of inspiration: the flutter text API (which somehow enables precise configuration while remaining rather ergonomic), and the semantics defined by the CSS specification. I'll take a deeper look to see if there are concepts we could use here..

benoitkugler:
Otherwise, I like the general plan

whereswaldon:
> Another possible parameter would be related to ellipsis (that is, add ... and truncate instead of wrapping).

I was thinking that we'd address this use-case by setting the `BreakWordPolicy` to always break the final word, setting `Hyphenate` to true, and setting `Hyphen` to a shaped ellipsis. I think that has the correct outcome, though I'll admit that it may not be super intuitive.

[...]

PR #37, opened 2022-12-14
shaping: implement simple line truncation

whereswaldon:
This commit adds a mechanism to limit the number of wrapped lines of text to a fixed quantity. Doing so will currently still require shaping all of the text, but does skip the work to wrap any truncated text.

Updates #28

This does not implement inserting a symbol to indicate truncation, but seemed like a good starting point. Being able to truncate without the symbol is useful on its own, and this way the change is as simple as possible.

I deviated slightly from the discussion in #28 because I realized there are reasons one might wish to only display exactly N lines of text. Instead of a boolean indicating whether or not to truncate, I've used an integer count of lines. I think it's reasonably intuitive, and has the happy side effect of having a zero value which will not perform truncation.

I look forward to your thoughts!

### DIFF
--- a/shaping/wrapping.go
+++ b/shaping/wrapping.go
@@ -314,10 +314,21 @@ type Range struct {
 // Output should be displayed sequentially on one line.
 type Line []Output
 
+// WrapConfig provides line-wrapper settings.
+type WrapConfig struct {
+	// TruncateAfterLines is the number of lines of text to allow before truncating
+	// the text. A value of zero means no limit.
+	TruncateAfterLines int
+}
+
 // LineWrapper holds reusable state for a line wrapping operation. Reusing
 // LineWrappers for multiple paragraphs should improve performance.
 type LineWrapper struct {
-	// seg is an internal storage used to initiate the breaker iterator
+	// config holds the current line wrapping settings.
+	config WrapConfig
+	// truncating tracks whether the wrapper should be performing truncation.
+	truncating bool
+	// seg is an internal storage used to initiate the breaker iterator.
 	seg segmenter.Segmenter
 
 	// breaker provides line-breaking candidates.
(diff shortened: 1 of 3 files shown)


benoitkugler:
Nicely done !
If I understand correctly, `WrapParagraph` will return at most WrapAfterLine (unless its zero), right? Maybe that could appear explicitly in the doc?

benoitkugler (approved):

shaping/wrapping.go
```diff
@@ -361,11 +374,11 @@ func (l *LineWrapper) Prepare(paragraph []rune, shapedRuns ...Output) {
 
 // WrapParagraph wraps the paragraph's shaped glyphs to a constant maxWidth.
 // It is equivalent to iteratively invoking WrapLine with a constant maxWidth.
-func (l *LineWrapper) WrapParagraph(maxWidth int, paragraph []rune, shapedRuns ...Output) []Line {
+func (l *LineWrapper) WrapParagraph(config WrapConfig, maxWidth int, paragraph []rune, shapedRuns ...Output) []Line {
```
whereswaldon, line 377:
```suggestion
// If the config has a non-zero TruncateAfterLines, WrapParagraph will return at most
// that many lines.
func (l *LineWrapper) WrapParagraph(config WrapConfig, maxWidth int, paragraph []rune, shapedRuns ...Output) []Line {
```

whereswaldon:
> Nicely done ! If I understand correctly, `WrapParagraph` will return at most WrapAfterLine (unless its zero), right? Maybe that could appear explicitly in the doc?

Done!

andydotxyz (approved):
Apologies for the delay, looks good.

whereswaldon:
Thanks all!

PR #40, opened 2023-01-07
Opentype (1/2)

benoitkugler:
This PR adds low level support for reading Opentype font files, in a new `opentype` folder.

The `opentype/loader` package only reads the file index, deferring the actual content parsing to the package `opentype/tables`.

A following PR will add an higher level API (used by Harfbuzz and equivalent to the current `textlayout/fonts.Face`), but I think this one is big enough.

Note that we now have a dep on `typesetting-utils`, but it should not impact end users thanks to module pruning.


benoitkugler:
>> Note that we now have a dep on typesetting-utils, but it should not impact end users thanks to module pruning.
> 
> I'm not sure that's true unless we declare `go 1.17` or higher in `go.mod`.

Ah, fair point... Should we do it ? in this PR ?

whereswaldon:
It kinda depends on how Go's forwards compatibility story works. I believe that @andydotxyz supports the most recent four versions of Go over in Fyne, so we need to support 1.1{9,8,7,6}. How does Go 1.16 handle building a project when the `go.mod` declares a higher number? I know that there is some measure in place to faciliate that, but I've never used it.

whereswaldon (approved):
I confess that I didn't read through the generated code, but I think this looks good. Thank you very much for the excellent work!

andydotxyz:
> It kinda depends on how Go's forwards compatibility story works. I believe that @andydotxyz supports the most recent four versions of Go over in Fyne, so we need to support 1.1{9,8,7,6}. How does Go 1.16 handle building a project when the `go.mod` declares a higher number? I know that there is some measure in place to faciliate that, but I've never used it.

Fyne still supports Go v1.14 so we can build 32bit iOS apps. We never move above Debian stable for version as we don't think developers should have to reach outside their package manager for a compiler version install. Later this year we will probably move to 1.16/17/18 TBD.

andydotxyz:
Reading https://go.dev/doc/modules/gomod-ref it sounds like that states you must have the version of the compiler as a minimum.

```
minimum-go-version
The minimum version of Go required to compile packages in this module.
```

(to compile Fyne with Go 1.14 you will have to have modules turned off so go-text/typesetting compiles). It will be nice to move to a mod-only-world in the future so our documentation can be simplified!

benoitkugler:
> Reading https://go.dev/doc/modules/gomod-ref it sounds like that states you must have the version of the compiler as a minimum.
> 
> ```
> minimum-go-version
> The minimum version of Go required to compile packages in this module.
> ```
> 
> (to compile Fyne with Go 1.14 you will have to have modules turned off so go-text/typesetting compiles). It will be nice to move to a mod-only-world in the future so our documentation can be simplified!

Thank you for the precision. 
What should we do then ? Is it OK to set typesetting go version to 1.17 so that users with a recent version can take advantage of module pruning ?

By the way, `benoitkugler/textlayout` already has go version as 1.17, and it does not seem problematic, does he ?

andydotxyz:
I did a little more testing and earlier versions of Go are able to build with 1.17 set. This is so strange.
I guess we would have known if this was not possible as our CI tests with 1.14 over the entire code base ;).

With that in mind I don't have a reason to hold it back. But I do worry about the mismatch of documented behaviour vs actual in the Go tooling!

whereswaldon:
I haven't been able to find the official docs on it, but I know that I've read building go versions that declare a higher go version than the current one is possible, so long as it isn't taking advantage of new language features.

benoitkugler:
> I did a little more testing and earlier versions of Go are able to build with 1.17 set. This is so strange. I guess we would have known if this was not possible as our CI tests with 1.14 over the entire code base ;).
> 
> With that in mind I don't have a reason to hold it back. But I do worry about the mismatch of documented behaviour vs actual in the Go tooling!

Thank your for the tests ! I'll send a separate PR to bump the go version then.

andydotxyz (approved):

PR #42, opened 2023-01-26
Opentype (2/2)

benoitkugler:
This PR adds an higher level API to use Opentype font files. It is meant to be

- consumed by Harbuzz during shaping
- used by the GUI toolkits when rendering glyphs

The structure of the package, with `api` and `api/font` is designed to support futur extension. For instance, a package fetching font metadata (footprints) could nicely be added as `api/metadata`; a sub-setter could go in `api/subset`.

(I've no strong opinion about the package names though.)

The `api/font` implementation exposes two types : a read-only `*Font`, which support concurrent uses and a light wrapper `Face`, so that toolkits may re-use `*Font` objects across the whole application.
Besides, the `Face` object is to be the direct replacement of the existing `textlayout/fonts.Face`.

Thank you for your suggestions !


opentype/api/font/font.go
```diff
@@ -0,0 +1,348 @@
...
+	cmapVar api.UnicodeVariations
+
+	hhea *tables.Hhea
+	vhea *tables.Vhea
+	vorg *tables.VORG // optional
+	cff  *cff.Font
+	post post // optional
+	svg  svg  // optional
+
+	// Optionnal, only present in variable fonts
```
sbinet, line 34:
```suggestion
	// Optional, only present in variable fonts
```

opentype/api/font/font.go
```diff
@@ -0,0 +1,348 @@
...
+	Kerx Kernx
+	GSUB GSUB // An absent table has a nil slice of lookups
+	GPOS GPOS // An absent table has a nil slice of lookups
+
+	head tables.Head
+
+	upem uint16 // cached value
+}
+
+// NewFont load all the font tables, sanitizing them.
```
sbinet, line 68:
```suggestion
// NewFont loads all the font tables, sanitizing them.
```

opentype/api/font/font.go
```diff
@@ -0,0 +1,348 @@
...
+	raw, _ = ld.RawTable(loader.MustNewTag("GSUB"))
+	layout, _, err := tables.ParseLayout(raw)
+	// harfbuzz rely on GSUB.Loookups being nil when the table is absent
+	if err == nil {
+		out.GSUB, _ = newGSUB(layout)
+	}
+
+	raw, _ = ld.RawTable(loader.MustNewTag("GPOS"))
+	layout, _, err = tables.ParseLayout(raw)
+	// harfbuzz rely on GSUB.Loookups being nil when the table is absent
```
sbinet, line 189:
```suggestion
	// harfbuzz rely on GPOS.Loookups being nil when the table is absent
```

opentype/api/font/font.go
```diff
@@ -0,0 +1,348 @@
...
+	if err == nil {
+		out.vorg = &vorg
+	}
+
+	// layout tables
+	out.GDEF, _ = loadGDEF(ld, len(out.fvar))
+
+	raw, _ = ld.RawTable(loader.MustNewTag("GSUB"))
+	layout, _, err := tables.ParseLayout(raw)
+	// harfbuzz rely on GSUB.Loookups being nil when the table is absent
```
sbinet, line 182:
```suggestion
	// harfbuzz relies on GSUB.Loookups being nil when the table is absent
```

opentype/api/font/cff/interpreter/interpreter.go
```diff
@@ -0,0 +1,363 @@
...
+	psArgStackSize = 48
+
+	// Similarly, Appendix B says "Subr nesting, stack limit 10".
+	psCallStackSize = 10
+
+	maxRealNumberStrLen = 64 // Maximum length in bytes of the "-123.456E-7" representation.
+)
+
+// PsContext is the flavour of the Postcript language.
+type PsContext uint32
```
sbinet, line 44: personally, I would reduce the stutter:
```suggestion
// Context is the flavour of the Postcript language.
type Context uint32
```

(as the name of the package already conveys the "postcript-ness" intent.)

opentype/api/font/cff/interpreter/interpreter.go
```diff
@@ -0,0 +1,363 @@
...
+		return errors.New("no subroutine has been called")
+	}
+	p.callStack.top--
+	// restore the previous instructions
+	p.instructions = p.callStack.vals[p.callStack.top]
+	return nil
+}
+
+// PsOperator is a postcript command, which may be escaped.
+type PsOperator struct {
```
sbinet, line 340: ditto:
```suggestion
// Operator is a postcript command, which may be escaped.
type Operator struct {
```

(as the name of the package already conveys the "postcript-ness" intent.)

opentype/api/font/cff/interpreter/interpreter.go
```diff
@@ -0,0 +1,363 @@
...
+
+func (p PsOperator) String() string {
+	if p.IsEscaped {
+		return fmt.Sprintf("2-byte operator (12 %d)", p.Operator)
+	}
+	return fmt.Sprintf("1-byte operator (%d)", p.Operator)
+}
+
+// PsOperatorHandler defines the behaviour of an operator.
+type PsOperatorHandler interface {
```
sbinet, line 353: ditto:
```suggestion
// OperatorHandler defines the behaviour of an operator.
type OperatorHandler interface {
```

(as the name of the package already conveys the "postcript-ness" intent.)

opentype/api/font/cff/interpreter/interpreter.go
```diff
@@ -0,0 +1,363 @@
...
+	// Context defines the precise behaviour of the interpreter,
+	// which has small nuances depending on the context.
+	Context() PsContext
+
+	// Apply implements the operator defined by `operator` (which is the second byte if `escaped` is true).
+	//
+	// Returning `ErrInterrupt` stop the parsing of the instructions, without reporting an error.
+	// It can be used as an optimization.
+	Apply(operator PsOperator, state *Machine) error
```
sbinet, line 362: usually, in Go APIs, it seems to me we try to put the modified argument "first":
```suggestion
	Apply(state *Machine, op Operator) error
```

opentype/api/font/cff/interpreter/interpreter.go
benoitkugler: Done.

opentype/api/font/cff/interpreter/interpreter.go
benoitkugler: Done.

opentype/api/font/cff/interpreter/interpreter.go
benoitkugler: Done.

opentype/api/font/cff/interpreter/interpreter.go
benoitkugler: Done.

benoitkugler:
@sbinet Thank you for you review, all your suggestions have been applied.

andydotxyz:
Thanks so much, am trying to grok it all but there is a lot here!

> * used by the GUI toolkits when rendering glyphs

Regarding this I was not sure what the main entry point is or how it would be used. Is one of the tests demoing the expected toolkit usage?

whereswaldon (approved):
I cannot provide extremely intelligent feedback here, as I lack the detailed knowledge of fonts. However, I think the exposed API looks reasonable. Thank you very much for all of the work you've put into this!

benoitkugler:
> Thanks so much, am trying to grok it all but there is a lot here!
> 
> > * used by the GUI toolkits when rendering glyphs
> 
> Regarding this I was not sure what the main entry point is or how it would be used. Is one of the tests demoing the expected toolkit usage?

The main entry point is the `Face.GlyphData` method, with a similar API to `textlayout`. For instance, you should be able to use it as a drop-in replacement in the new go-text/render implementation.

There are basic tests in renderer_test.go, but no complete example (as I feel like it is outside the scope of the package).

andydotxyz (approved):

benoitkugler:
Great, thanks !

PR #44, opened 2023-03-21
Wrapping Refactors

whereswaldon:
I'm working to implement the insertion of a text truncation symbol, and I've found the wrapping algorithm to be quite difficult to read (it's long, it's intricate, etc). These commits are an effort to make the core algorithm clearer, and to provide reusable helpers. There's definitely more that can be done here, but I thought these changes made a dent, and are probably not contentious.


benoitkugler (approved):
Looks good to me !

andydotxyz (approved):
Nice improvement

PR #45, opened 2023-03-21
ci: update to latest staticcheck

whereswaldon:
Just a version bump for the linter and some adaptations to newer lint rules.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -17,5 +17,5 @@ jobs:
 
     - name: Staticcheck
       run: |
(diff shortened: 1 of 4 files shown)


andydotxyz:
> > Looks good. I am curious why those lint ignores are not needed now though.
> 
> I suspect that they're only enforced when your go.mod declares a sufficiently high version of go. Since we're striving to be backwards compatible pretty far, staticcheck is probably disabling them for us.

Oh I like it, that's smart :)

benoitkugler:
Thanks for the cleanup !

benoitkugler (approved):

PR #46, opened 2023-03-22
Truncator support

whereswaldon:
This commit builds upon #44 to actually implement a text truncator symbol.

Some design choices:

- I tried to implement what we discussed in #28 towards the end of the thread.
- The `Truncator` is a single `Output`. This limits it to being a single font face, size, script, and language, though it can contain any number of glyphs. Supporting a `[]Output` truncator is a solid complexity hike, and I'm not sure that they are really used in practice.
- To make it easier to populate the `Truncator` field, I've provided a helper on `WrapConfig` that accepts a `Shaper` and `Input`.
- Applications often want to know how much of the string was truncated. I've made this an additional return value from the wrapping functions because I didn't see a clearer way to communicate it right now.

Only the final commit in this PR is new. Please focus review on that and review the rest on #44

### DIFF
--- a/shaping/fuzz_test.go
+++ b/shaping/fuzz_test.go
@@ -33,7 +33,7 @@ func FuzzE2E(f *testing.F) {
 			t.Errorf("expected %d shaped runes, got %d", len(textInput), out.Runes.Count)
 		}
 		var l LineWrapper
-		outs := l.WrapParagraph(WrapConfig{}, 100, textInput, out)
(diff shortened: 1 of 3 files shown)


benoitkugler (approved):
This looks good to me ! Very nice work !
I agree that a single output for the truncator seems enough in practice.

andydotxyz (approved):
Sweet :)

PR #48, opened 2023-03-25
Fix leak if fonts are replaced in usage

### LINKED ISSUE #47: Shaper is leaking font data
If I load a font, then shape it, then discard the font it will remain in memory.
For an app that always uses the same font this is not a problem, but if switching data a lot this will grow in memory very fast!

For example, watch this code use 1GB in under a minute:

```go
package main

import (
	"bytes"
	"io/ioutil"
	"time"

	"github.com/go-text/typesetting/di"
	"github.com/go-text/typesetting/font"
	"github.com/go-text/typesetting/shaping"
)

func main() {
	data, _ := ioutil.ReadFile("../../theme/font/NotoSans-Regular.ttf")

	for range time.Tick(time.Millisecond * 100) {
		f, _ := font.ParseTTF(bytes.NewReader(data))

		in := shaping.Input{
			Text:      []rune{' '},
			RunStart:  0,
			RunEnd:    1,
			Direction: di.DirectionLTR,
			Face:      f,
			Size:      14,
		}
		shaper := &shaping.HarfbuzzShaper{}
		_ = shaper.Shape(in) // this line is leaking the font!
	}
}
```

andydotxyz:
Move plan cache to the buffer object so we don't leak if the shaper / font is not needed any more

Fixes #47

### DIFF
--- a/harfbuzz/buffer.go
+++ b/harfbuzz/buffer.go
@@ -105,8 +105,7 @@ type Buffer struct {
 
 	haveOutput bool
 
-	planCache     map[Face][]*shapePlan
-	planCacheLock *sync.Mutex
+	planCache map[Face][]*shapePlan
 }
 
 // NewBuffer allocate a storage with default options.
@@ -116,7 +115,6 @@ func NewBuffer() *Buffer {
 		ClusterLevel:  MonotoneGraphemes,
 		maxOps:        maxOpsDefault,
 		planCache:     map[Face][]*shapePlan{},
-		planCacheLock: &sync.Mutex{},
 	}
 }
 
--- a/harfbuzz/buffer.go
+++ b/harfbuzz/buffer.go
@@ -2,6 +2,7 @@ package harfbuzz
 
 import (
 	"math"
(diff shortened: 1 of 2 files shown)


benoitkugler (changes requested):
This seems fine to me, expect that we should not add a lock since the `Buffer` type is *not* safe for concurrent use anyway.

benoitkugler:
As a side note, I think it could make sense for Fyne to only create one Font object per byte slice, since it is a rather costly operation (like Gioui does).
That would only reduce the leak, not remove it though, so this PR seems still required ;)

andydotxyz:
> As a side note, I think it could make sense for Fyne to only create one Font object per byte slice, since it is a rather costly operation (like Gioui does).

I agree, I thought Fyne basically did the same. The leak was when changing theme at which point we do need to read new font data.

benoitkugler (approved):
Thanks, looks good to me !

whereswaldon (approved):
Makes sense. Thanks!

Issue #49, opened 2023-03-26
Font API : collections, concurrency


benoitkugler:
We should polish our API for font loading : the heavy lifting is already implemented in `opentype/api/font` but some features are missing in the public `font` package.

Firstly, we should export a function to support collections. What do you think of 
```go
ParseTTC(Resource) ([]Face, error)
```
?

On the other hand, concurrent usage of fonts was something @whereswaldon was interested in. It is now possible, but requires to properly split the font usage in two parts : the `*opentype/api/font.Font` which is safe for concurrent use, and the `*font.Face`, which is not. Should we clarify that in the `font` package, for instance by also exporting the `*opentype/api/font.Font` type ?

What do you think overall about font handling ?

whereswaldon:
I'd love support for collections again. That was a feature Gio lost when switching to our shaper.

That signature looks fine to me.

I would like to expose the dual font/face types so that I could build a shared font loading cache across all of Gio's shaper instances, with each shaper building faces for its own use.

One challenge we have in Gio right now is attaching correct metadata to loaded fonts. We currently kinda force users to do it manually for custom fonts that they load. Is there any consistent way to extract that information from the font data itself? The name of the typeface, the style (bold, italic/emphasis), etc...? Or is that somehow inferred by other software (say, web browsers)?

benoitkugler:
> One challenge we have in Gio right now is attaching correct metadata to loaded fonts. We currently kinda force users to do it manually for custom fonts that they load. Is there any consistent way to extract that information from the font data itself? The name of the typeface, the style (bold, italic/emphasis), etc...? Or is that somehow inferred by other software (say, web browsers)?

Yes, there is a way : on linux systems, this is typically done by fontconfig. I've actually written somewhat similar (but simplified) in the [font-scan](https://github.com/go-text/typesetting/blob/font-scan/fontscan/properties.go
) branch.
There are 4 properties available : 

- family name
- weight (bold, ...)
- style (italic, ...)
- stretch (condensed, ...)

This is something I will eventually need as well when using go-text for html rendering, so I can propose a PR to integrate the relevant part in go-text : that would be a in new package `opentype/api/metadata/`

whereswaldon:
I was actually looking at the `font-scan` branch recently because I'm thinking about how to properly do system font loading for Gio in the future. I think the approach definitely has merit.

I'd happily review a new PR for the metadata API. :D In the longer term, we should rebase the font-scan branch and open a PR for that functionality as well.

[...]

PR #51, opened 2023-03-28
Add an API to query font metadata such as family name and style

benoitkugler:
This PR provides a way to query metadata about a font (see also #49)

The changes are large because some code has moved to the `tables` package to avoid duplication.

The various constants are taken from the CSS specifications (also used by other librairies like font-kit).

For now, the API requires a `*loader.Loader`,  meaning the original font file is needed. This is required since the `*font.Font` type does not store all the metadata, it only selects the shaping/glyph related information.

### DIFF
--- a/opentype/api/cmap.go
+++ b/opentype/api/cmap.go
@@ -81,57 +81,37 @@ func ProcessCmap(cmap tables.Cmap) (Cmap, UnicodeVariations, error) {
 	}
 
 	// now find the best cmap, following harfbuzz/src/hb-ot-cmap-table.hh
-	const (
-		PlatformUnicode tables.PlatformID = iota
-		PlatformMac
-		PlatformIso /* deprecated */
-		PlatformMicrosoft
-		PlatformCustom
-		_
-		_
-		PlatformAdobe /* artificial */
-	)
-	const (
-		PEUnicodeDefault     = tables.EncodingID(0)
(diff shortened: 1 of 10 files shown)


whereswaldon (commented):
Only one very minor recommendation about how constants are defined. Otherwise this looks fantastic! Is there a specific reason that you marked it as a draft?

benoitkugler:
> Only one very minor recommendation about how constants are defined. Otherwise this looks fantastic! Is there a specific reason that you marked it as a draft?

Hum no, not really, just because I was not sure about the constants or the `Metadata` signature.

whereswaldon (approved):
This all looks good to me. Thank you very much! Should we offer a way to expose this in `typesetting/font`? Maybe `ParseTTFWithMetadata`? We certainly don't need to, as applications can implement that themselves. It's more a question of what we want to offer out of the box.

benoitkugler:
> This all looks good to me. Thank you very much! Should we offer a way to expose this in `typesetting/font`? Maybe `ParseTTFWithMetadata`? We certainly don't need to, as applications can implement that themselves. It's more a question of what we want to offer out of the box.

We could maybe be a bit conservative here and wait before adding it ? We will probably have other metadata related features in the future.

andydotxyz (approved):

PR #52, opened 2023-03-28
shaping: allow forcing truncation

whereswaldon:
This commit enables application code to require the truncator to be inserted when the text wraps to the maximum number of lines even when the text would normally fit without truncation. Our API is paragraph-centric, and thus does not know about paragraphs after itself. If trying to display truncatable text with multiple paragraphs of contents, the previous API could not display a truncator if the text wrapped a non-final paragraph to exactly fill the maximum number of lines. This would result in the text not appearing to be truncated, while it is missing entire lines of content.

The only reasonable solution I could see was to offer the option to always insert the truncator when wrapping the final line. Application code can then track whether there are future paragraphs (via readahead) and choose to request that the current paragraph display as truncated if there are future paragraphs and this paragraph fills all available lines.

### DIFF
--- a/shaping/wrapping.go
+++ b/shaping/wrapping.go
@@ -322,6 +322,15 @@ type WrapConfig struct {
 	// Truncator, if provided, will be inserted at the end of a truncated line. This
 	// feature is only active if TruncateAfterLines is nonzero.
 	Truncator Output
+	// TextContinues indicates that the paragraph wrapped by this config is not the
+	// final paragraph in the text. This alters text truncation when filling the
+	// final line permitted by TruncateAfterLines. If the text of this paragraph
+	// does fit entirely on TruncateAfterLines, normally the truncator symbol would
+	// not be inserted. However, if the overall body of text continues beyond this
+	// paragraph (indicated by TextContinues), the truncator should still be inserted
+	// to indicate that further paragraphs of text were truncated. This field has
+	// no effect if TruncateAfterLines is zero.
+	TextContinues bool
 }
 
 // WithTruncator returns a copy of WrapConfig with the Truncator field set to the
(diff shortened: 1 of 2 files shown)


whereswaldon:
An example: imagine truncating the string `"hi\nthere"` to a maximum of one line. You have to break it into two different shapings, as paragraphs are the unit of shaping. So you shape `"hi\n"` and it doesn't get truncated, so you don't show an ellipsis. Then you don't want to shape `"there"` because you know you've used all of the lines already.

It's tempting to say "just append a truncator", but that won't work if the first line just barely fits, an edge case handled by special logic in the line wrapper. Basically you need a way to say "if we hit the final line while wrapping this paragraph, ensure the truncator is inserted even if all of the text fits"

You, the application programmer, would do this when you know that there are subsequent paragraphs of text that might be truncated when this paragraph is.

andydotxyz:
Regarding the naming I just wonder if it should align more toward the generic "there is more text" meaning then instead of the more specific "force truncation"? 
If there are other uses in the future this could offer more flexibility? I.e. “TextContinues” instead of “ForceTruncation”

I'm not sure if this is the right thing to do but wanted to raise the question.

whereswaldon:
> Regarding the naming I just wonder if it should align more toward the generic "there is more text" meaning then instead of the more specific "force truncation"? If there are other uses in the future this could offer more flexibility? I.e. “TextContinues” instead of “ForceTruncation”

It's a good question. If there were future wrapping features that needed to know whether we were on the final paragraph, they could benefit from such a rename. However, that would only make sense if they should *always* be activated based on whether we were on the final paragraph.

> I'm not sure if this is the right thing to do but wanted to raise the question.

I'm honestly happy to rename this. I don't know how likely it is that we'll have more features that want it, but my level of care about the name is quite small. I just want the feature so that we can correctly truncate multi-paragraph text reliably.

whereswaldon:
I've pushed a revision renaming the field and trying to improve the documentation for it.

benoitkugler:
I like the renaming, I think its explains the problem and the solution a bit better ! Thanks the clearer documentation as well !

benoitkugler (approved):

andydotxyz (approved):
Thanks, I think this reads very cleanly

Issue #53, opened 2023-03-29
High memory allocation


Bluebugs:
Harfbuzz shaper Shape() function does generate a large amount of memory allocation due to the call to harfbuzz.NewFont() at : https://github.com/go-text/typesetting/blob/main/shaping/shaper.go#L77 . I have done a quick/hacky local experiment by putting that font in a map cache and the resulting operation save significant amount of memory allocation and seems to also improve performance significantly.

Does anyone have an idea on what would be the proper approach to cache this font structure?

whereswaldon:
@benoitkugler Can that type be reused? If so, it does seem like it would be worthwhile to add tunable LRU caching for it.

PR #54, opened 2023-03-31
Cache harfbuzz fonts

benoitkugler:
This change (for #53) leverages the fact that the `HarfbuzzShaper` operates sequentially, so that modifying a `harbuzz.Font` is safe.
(It wouldn't be the case with concurrent use, since a same `*font.Font` may be used with different user provided settings).

A test/benchmark would be welcomed :)

### DIFF
--- a/shaping/shaper.go
+++ b/shaping/shaper.go
@@ -5,7 +5,6 @@ package shaping
 import (
 	"github.com/go-text/typesetting/di"
 	"github.com/go-text/typesetting/harfbuzz"
-	"github.com/go-text/typesetting/opentype/api/font"
 	"golang.org/x/image/math/fixed"
 )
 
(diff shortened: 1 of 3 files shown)


whereswaldon (changes requested):
I like the direction this is taking, but I have a resource use concern.

shaping/shaper.go
```diff
@@ -14,6 +15,8 @@ import (
 // for each operation.
 type HarfbuzzShaper struct {
 	buf *harfbuzz.Buffer
+
+	fonts map[*font.Font]*harfbuzz.Font
```
whereswaldon, line 19: I think we need to make this an LRU. As it stands, there's no upper bound on the quantity of memory this will take if you use the same `HarfbuzzShaper` for a long time within a single application.

benoitkugler: What should be the capacity of the cache ?

whereswaldon:
I've opened a suggestion PR against this branch: https://github.com/go-text/typesetting/pull/56

Let me know what you think. :D

whereswaldon (approved):
All of my concerns are addressed.

andydotxyz (approved):
Nice one, thanks for doing this

PR #55, opened 2023-03-31
[font] export support for font collections and add alias for font.Font

benoitkugler:
See #49 

I've merged the two files in the [font] package : since we won't add much in this package, two files seems now a bit of an overkill.


whereswaldon (approved):
Thanks! It should be possible to write much more efficient APIs on top of the newly-exposed stuff here.

PR #59, opened 2023-04-11
Reuse errors in functions that get called a lot

dominikh:
sbix.glyphData and bitmap.glyphData get called repeatedly even for fonts that have neither. Previously, they allocated an error on each call.


dominikh:
From a profile:
image

benoitkugler (approved):
Nice optimisation, thank you !

andydotxyz (approved):

PR #62, opened 2023-04-12
Test agreed spread of versions

andydotxyz:
I think we didn't apply this after our chat on Slack


benoitkugler (approved):
Sounds good to me, thanks

whereswaldon (approved):
Looks good now, thank you!

PR #63, opened 2023-04-23
Font scan

benoitkugler:
As discussed in #17, I've rebased this branch to include the latest changes.

The PR is long : you may want to start by the [readme](https://github.com/go-text/typesetting/tree/font-scan/fontscan) before diving into the implementation.

As a summary, this package provides two functionalities : 

- a way to scan the file system and extract metadata about font files 
- given a list of font metadata, the logic to select the proper one given a family, style and rune query (implementing the CSS spec about font selection)

I'm open to all your suggestions !


whereswaldon:
I haven't yet been able to dig into the implementation, but I've read the README and it sparked some thoughts:

- I think the current FindFont strategy is unable to load mono fonts, as that info isn't present in the family/aspect provided. Furthermore, it's more expensive to check that. I think this would be a common UI toolkit need though. GUI authors want to be able to resolve a system monospace font quickly.
- I'm starting to second-guess whether the fontmap approach might not be suitable for GUI applications. Loading a font by rune coverage is a common operation. If the initial scan for system fonts is only a second most of the time and subsequent launches reuse the work, that might be viable after all.

I'll wade into the implementation some more soon.

fontscan/fontmap.go
```diff
@@ -0,0 +1,394 @@
...
+
+		// load an existing index
+		var execPath string
+		execPath, err = os.Executable()
+		if err != nil {
+			err = fmt.Errorf("resolving index cache path: %s", err)
+			return
+		}
+
+		cachePath := filepath.Join(filepath.Dir(execPath), cacheFile)
```
whereswaldon, line 122: This isn't a safe way to determine where to store the cache. In macOS application bundles, for example, the directory containing the executables is readonly. On Linux, it would be highly unusual to find `font_index.cache` in a `bin` directory, and some application sandboxing may prevent this outright.

Every OS has an application-writable configuration directory though. On Linux it's usually `$XDG_CONFIG_HOME/application`, on macOS it's `/Users/me/Library/Application Support/application`, and on Windows it's `C:\User\me\AppData\Local\application`. Android and iOS are compatible (I think) with the Linux and macOS strategies, which only leaves WASM as the tricky case. There's probably a way to use local storage for this in WASM.

I think we discussed this in the past as well, but if we were able to make the font cache format backwards and forwards compatible, it would be possible to share the cache across applications. Imagine Fyne and Gio applications being able to leverage the same system font cache on the same computer, rather than each generating their own! Then we could place it somewhere generic (like a folder called `gotext`) instead of in a per-application location.

fontscan/substitutions.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+}
+
+// we want to easily insert at the start,
+// the end and "around" an element
+type familyList struct {
+	*list.List
+}
+
+// panic if `families` is empty
```
whereswaldon, line 39: This doesn't look like it actually will panic. Old docs?

fontscan/substitutions.go
```diff
@@ -0,0 +1,254 @@
...
+// must equals `lang`
+type langAndFamilyEqual struct {
+	lang   language.Language
+	family string
+}
+
+// TODO: for now, these tests language base tests are ignored
+func (langAndFamilyEqual) test(list familyList) *list.Element {
+	return nil
+}
```
whereswaldon, line 187: What's the impact of leaving this unimplemented? We won't be able to correctly apply substitution predicates that require a particular family and language combination?

fontscan/substitutions.go
```diff
@@ -0,0 +1,254 @@
...
+// must contains `lang`
+type langContainsAndFamilyEquals struct {
+	lang   language.Language
+	family string
+}
+
+// TODO: for now, these tests language base tests are ignored
+func (langContainsAndFamilyEquals) test(list familyList) *list.Element {
+	return nil
+}
```
whereswaldon, line 204: Similarly, leaving this unimplemented prevents us from applying a substitution that has a set of possible languages and a family requirement?

whereswaldon (commented):
Wow, what a huge chunk of work @benoitkugler. As always, thank you so much for your dedication. I understand what is going on here reasonably well after my initial review, and I'm excited about it. My next step is to try to build a branch of Gio that uses this in order to find any API issues that make this hard to consume from the toolkit side.

fontscan/rune_coverage.go
```diff
@@ -0,0 +1,267 @@
...
+//
+// The internal representation is a slice of `pageSet` pages, where each page is a boolean
+// set of size 256, encoding the last byte of a rune.
+// Each rune is then mapped to a page index (`pageNumber`), defined by it second and third bytes.
+
+// pageSet is the base storage for a compact rune set.
+// A rune is first reduced to its lower byte 'b'. Then the index
+// of 'b' in the page is given by the 3 high bits (from 0 to 7)
+// and the position in the resulting uint32 is given by the 5 lower bits (from 0 to 31)
+type pageSet [8]uint32
```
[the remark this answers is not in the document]
benoitkugler: All credits to fontconfig authors ! But yes, bit set are pretty amazing :)

fontscan/substitutions.go
benoitkugler: Indeed, thanks !

fontscan/substitutions.go
```diff
@@ -0,0 +1,254 @@
...
+// must equals `lang`
+type langEqualsAndNoFamily struct {
+	lang   language.Language
+	family string
+}
+
+// TODO: for now, these tests language base tests are ignored
+func (langEqualsAndNoFamily) test(list familyList) *list.Element {
+	return nil
+}
```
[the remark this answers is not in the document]
benoitkugler: This is slightly different here : no family means that the family set in the test must no appear in the current family list (queried by the user, including substitutions).
For instance the test `langEqualsAndNoFamily{lang: "und-zmth", family: "math"}` means : for math language (this is the special tag `und-zmth`), if the family 'math' is _not_ queried, do ....
This test would trigger on (given "und-zmth" is set)  `{"arial", "monospace"}`, but not on `{"arial", "math", "monospace"}`.

fontscan/fontmap.go
benoitkugler: Fair enough, thank you !
I've extracted the cache directory in `cacheDir` and used `os.UserConfigDir` as first implementation. 
Feel free to update it :) In particular, I'm not sure this works on mobile...
For WASM, we would need `syscall/js` right ? like in [this library](https://github.com/jonhteper/wasm/blob/main/storage.go) ?

whereswaldon: Yeah, we'd need `syscall/js` for sure. However, I don't know how we would go about discovering available fonts. Seems like it'd be a big effort. We could initially just make the system font scan return an empty set when running in the browswer. Then we could expand after someone has time to investigate how to do the operation correctly within that runtime environment.

andydotxyz: > Android and iOS are compatible (I think) with the Linux and macOS strategies, which only leaves WASM as the tricky case. There's probably a way to use local storage for this in WASM.

I don't think that is accurate for iOS - the cache location will be in the app sandbox not in the `$USER/Library/...` location won't it?

> I think we discussed this in the past as well, but if we were able to make the font cache format backwards and forwards compatible, it would be possible to share the cache across applications.

We can't do that for mobile due to app sandbox issues I think.

> We could initially just make the system font scan return an empty set when running in the browswer.

I'd support this on a first pass. System lookups from a web browser is likely a big topic!

andydotxyz:
What a phenomenal effort here.
I think as well as testing from an API perspective we need to get it compiled and running on iOS/Android devices to check we're not stepping out of sandboxes etc - will likely cause panics or ineffective caches if we do.

fontscan/fontmap.go
benoitkugler: @whereswaldon Sure thing, thank you for digging this up. Perhaps you could edit `DefaultFontDirectories` in `scan.go` accordingly ?

whereswaldon: I'll be happy to. I've done a local rebase of this atop latest `main`, which I plan to put up as a separate draft PR so you can double-check my work. I'll include this change there.

benoitkugler:
> * The font querying needs to support all of the metadata fields. Right now it doesn't support querying by `IsMonospace`, which is important for Gio's use cases. I'm relatively confident that this is easy to fix.

The logic for now is that monospace fonts are specified via families : the "monospace" special family name is replaced by suitable families. However, we may indeed include a `IsMonospace` flag in the footprint. (Besides, it could probably be more reliable since it would be computed from the actual glyph widths.)

benoitkugler:
> * I had to hack the fontmap to return both the font and its metadata every time I resolved a face. Gio always needs the metadata for a font when working with it. We track fonts by their metadata internally, and without it we have no key for the font value. To make the FontMap work seamlessly with Gio, I'd either need the API to return the metadata like my hacked version, or for fonts to generally always carry their metadata (I've had to hack around the lack of it previously). I'm not sure what the tradeoffs are there, beyond a slight increase in the resource requirements to load a font. @benoitkugler Your insight here would be appreciated.

Hum... I thought the (absolute) file name of a font was enough as a key (well the `Location` type) : do you also need the metadata after having resolved the face, or do you only use it as a key ?

But we could indeed return the metadata: after all the `Footprint` is a super set of the `Metadata` struct

whereswaldon:
> > Of the problems, the fontconfig seems like the biggest. Good Linux support is near and dear to my heart, and I want this to work well. At a high level, I think we have two options to make this work:
> 
> I think I prefer the first option, because, in my use case, I would rather avoid using fontconfig, so that the font selection logic is uniform across systems.
>
> Ignoring user specified settings for now, it seems that it should be enough to implement the first item :
> 
> > parsing fontconfig's configuration files to extract system font directories
> 
> since we can then scan these directories and update our index if necessary. Am I correct ? If so, I think is it relatively easy : in my port, I've written a fontconfig XML format parser, and I could simplify it to just fetch the font directories. This preliminary step could be added in `DefaultFontDirectories` (which is called at each app startup).

Yes, I think this would be sufficient. I had forgotten that we re-scan the timestamps of the fonts on startup. Hopefully we can do this whole process pretty quickly.

> By the way,
> 
> > parsing the actual contents of fontconfig caches and building our cache type out of them
> 
> would be quite hard because it would require to understand the in-memory layout used by fontconfig types, which is doable put a bit painful and error prone. So I would rather do this as a last solution.

Ah, so it's unpacking that directly into C structures? In that case, I totally agree. That's a nightmare.

benoitkugler:
So, to sum up and plan our next steps, we should at least tackle the following points : 

- [ ] Add `IsMonospace` to the footprint and the query
- [ ] Return the metadata as well as the face 
- [ ] On linux, scan fontconfig config file to detect font directories

I'll start with point 3 later.

whereswaldon:
My next steps are figuring out the failing tests on my system. Then I can pick up the next item from @benoitkugler's checklist.

benoitkugler:
> I've figured out one of my test failures. `TestScanIncrementalUpdate` was failing because my `ext4` filesystem [may not be able to provide nanosecond resolution modification times depending upon my kernel config](https://stackoverflow.com/a/14393315). Inserting a synthetic sleep between the various modifications makes the test pass reliably, but raises a question: does this sort of imprecision in the filesystem metadata have implications for the modification time tracking as a whole?

Interesting !
I think in practice the imprecision is not a big deal, since two scans would typically run with a large time delta between them (for now, two application startups)

> There's also some question about how this intersects with OS packaging. Some package managers install their content files with the modification timestamps from the contents within the package. We don't actually have a guarantee that a file will have a different timestamp from one package version to the next.

Would it be possible that two files with the same path have different content but the same timestamp ?

benoitkugler:
> My other local test failure is:
> 
> ```
> --- FAIL: TestResolveFont (0.16s)
>     fontmap_test.go:43: unexpected logs 2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
>         2023/06/14 13:27:14 No font matched for [helvetica] -> returning arbitrary face
> ```
> 
> This test case uses `UseSystemFonts()` and assumes that something will always match helvetica, but apparently my system doesn't have any candidates.

Very interesting failure : the Helvetica family is known by the library, so family substitutions should trigger and be enough. More precisely, for Helvetica, the following families are selected : 
[helvetica nimbussans nimbussansl texgyreheros arial arimo liberationsans albany albanyamt dejavulgcsans dejavusans bitstreamverasans verdana luxisans lucidasansunicode bpgglahointernational tahoma urwgothic nimbussansnarrow loma waree garuda umpush laksaman notosanscjkjp notosanscjkkr notosanscjksc notosanscjktc notosanscjkhk lohitdevanagari droidsansfallback khmeros nachlieli yuditunicode kerkis armnethelvetica artsounk bpgutf8m saysetthaunicode jglaooldarial gfzemenunicode pigiarniq bdavat bcompset kacst-qr urdunastaliqunicode raghindi muktinarrow padmaa hapaxberbère msgothic umepluspgothic microsoftyahei microsoftjhenghei wenquanyizenhei wenquanyibitmapsong arplshanheisununi arplnewsung mgopenmoderna mgopenmodata mgopencosmetica vlgothic ipamonagothic ipagothic sazanamigothic kochigothic arplkaitimgb arplkaitimbig5 arplsungtilgb arplmingti2lbig5 ｍｓゴシック zysong18030 nanumgothic undotum baekmukdotum baekmukgulim kacstqura lohitbengali lohitgujarati lohithindi lohitmarathi lohitmaithili lohitkash
miri lohitkonkani lohitnepali lohitsindhi lohitpunjabi lohittamil meera lohitmalayalam lohitkannada lohittelugu lohitoriya lklug freesans arialunicodems arialunicode code2000 code2001 sans-serif roya koodak terafik itcavantgardegothic helveticanarrow]
Could you manually check that none of these fonts are present on your system ? That seems strange to me... maybe ArchLinux does not come with a lot of fonts installed ?

whereswaldon:
> > This test case uses `UseSystemFonts()` and assumes that something will always match helvetica, but apparently my system doesn't have any candidates.
> 
> Very interesting failure : the Helvetica family is known by the library, so family substitutions should trigger and be enough. More precisely, for Helvetica, the following families are selected : [helvetica nimbussans nimbussansl texgyreheros arial arimo liberationsans albany albanyamt dejavulgcsans dejavusans bitstreamverasans verdana luxisans lucidasansunicode bpgglahointernational tahoma urwgothic nimbussansnarrow loma waree garuda umpush laksaman notosanscjkjp notosanscjkkr notosanscjksc notosanscjktc notosanscjkhk lohitdevanagari droidsansfallback khmeros nachlieli yuditunicode kerkis armnethelvetica artsounk bpgutf8m saysetthaunicode jglaooldarial gfzemenunicode pigiarniq bdavat bcompset kacst-qr urdunastaliqunicode raghindi muktinarrow padmaa hapaxberbère msgothic umepluspgothic microsoftyahei microsoftjhenghei wenquanyizenhei wenquanyibitmapsong arplshanheisununi arplnewsung mgopenmoderna mgopenmodata mgopencosmetica vlgothic ipamonagothic ipagothic sazanamigothic kochigothic arplkaitimgb arplkaitimbig5 arplsungtilgb arplmingti2lbig5 ｍｓゴシック zysong18030 nanumgothic undotum baekmukdotum baekmukgulim kacstqura lohitbengali lohitgujarati lohithindi lohitmarathi lohitmaithili lohitkash miri lohitkonkani lohitnepali lohitsindhi lohitpunjabi lohittamil meera lohitmalayalam lohitkannada lohittelugu lohitoriya lklug freesans arialunicodems arialunicode code2000 code2001 sans-serif roya koodak terafik itcavantgardegothic helveticanarrow] Could you manually check that none of these fonts are present on your system ? That seems strange to me... maybe ArchLinux does not come with a lot of fonts installed ?

Arch Linux comes with no fonts installed, but I have quite a few on the system now. :D I definitely have Liberation Sans.

whereswaldon:
I realized that the CI config was messed up during the rebase and fixed it. Looks like we can't use `io/fs` in `fontscan` because it wasn't introduced until something like Go 1.16

benoitkugler:
> Arch Linux comes with no fonts installed, but I have quite a few on the system now. :D I definitely have Liberation Sans.

I guess the next steps would be to check if `UseSystemFonts` have properly scanned and added LiberationSans, and then to check what the content of `Fontmap.candidates` is after calling `SetQuery`

whereswaldon:
> > Arch Linux comes with no fonts installed, but I have quite a few on the system now. :D I definitely have Liberation Sans.
> 
> I guess the next steps would be to check if `UseSystemFonts` have properly scanned and added LiberationSans, and then to check what the content of `Fontmap.candidates` is after calling `SetQuery`

I've been crawling through the test in [rr](https://rr-project.org/), and I think it's actually failing to resolve a face for `تثذرزسشص`, the Latin test text is fine. What's odd about that is that I can see this text in my browser, so I have a font somewhere that can display it. However, my terminal emulator can't display these characters... Makes me wonder if that font is in a nonstandard location or something.

benoitkugler:
I've pushed 7ef5a64 with support for fontconfig font directories. Let me know what you think. 
I was not sure if openbsd and freebsd use fontconfig as well. If so, we should also update `DefaultFontDirectories`.

benoitkugler:
> I realized that the CI config was messed up during the rebase and fixed it. Looks like we can't use `io/fs` in `fontscan` because it wasn't introduced until something like Go 1.16

Since Fyne will target go 1.17 in its next release, we could maybe wait for it and bump the tests to 1.16 before merging font-scan.

whereswaldon:
> > I realized that the CI config was messed up during the rebase and fixed it. Looks like we can't use `io/fs` in `fontscan` because it wasn't introduced until something like Go 1.16
> 
> Since Fyne will target go 1.17 in its next release, we could maybe wait for it and bump the tests to 1.16 before merging font-scan.

It's not a problem in the test code, but in the actual fontscan implementation. We don't run tests on 1.14, we just try to compile.

In general, I'd rather not wait for a Fyne release. I'll investigate what it would take to remove that dependency from the fontscan implementation.

fontscan/fontconfig.go
```diff
@@ -0,0 +1,178 @@
...
+				}
+			}
+			includes = append(includes, include)
+		}
+	}
+	return
+}
+
+// parseFcDir processes all the files in [dir] matching the [09]*.conf pattern
+// seen is updated with the processed fontconfig files
```
whereswaldon, line 106: I don't think that using a numeric prefix is actually *required* for those files. Users can insert new ones that don't follow the pattern, they just may have unexpected ordering due to lexicographic sorting. We should parse anything with the right extension.

benoitkugler: According to fontconfig documentation : 

> If a directory, every file within that directory starting with an ASCII digit (U+0030 - U+0039) and ending with the string ``.conf'' will be processed in sorted order. 

Note this is only true for _directories_, not for single files.

benoitkugler:
> It's not a problem in the test code, but in the actual fontscan implementation. We don't run tests on 1.14, we just try to compile.
> 
> In general, I'd rather not wait for a Fyne release. I'll investigate what it would take to remove that dependency from the fontscan implementation.

Thank you for the tedious backport code..

whereswaldon:
I've been pouring over [the fontconfig docs](https://www.freedesktop.org/software/fontconfig/fontconfig-user.html) and [the XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) to try to figure out exactly what we need to load and in what order so that we're reliably getting everything and actually loading config files in the user's home directory.

Users frequently have font config in `~/.local/share/fonts`, `~/.config/fontconfig` and maybe other places. I wanted to make sure we didn't miss anything.

XDG says:

> A specification that refers to $XDG_DATA_DIRS or $XDG_CONFIG_DIRS should define what the behaviour must be when a file is located under multiple base directories. It could, for example, define that only the file under the most important base directory should be used or, as another example, it could define rules for merging the information from the different files. 

So basically it's up to fontconfig to decide which of these config files takes precedence.

Fontconfig says:

> $XDG_CONFIG_HOME/fontconfig/conf.d and ~/.fonts.conf.d is the conventional name for a per-user directory of (typically auto-generated) configuration files, although the actual location is specified in the global fonts.conf file.

> $XDG_CONFIG_HOME/fontconfig/fonts.conf and ~/.fonts.conf is the conventional location for per-user font configuration, although the actual location is specified in the global fonts.conf file.

Which implies that the global config file is always responsible for loading the user-specific ones.

The only remaining question is how to correctly locate the global fontconfig file? I've tried to figure out the process below:

1. We start by examining `$FONTCONFIG_FILE`. If that is set, it's the location of the root XML config file. If it isn't set, assume it's `fonts.conf`. Resolve that file's location (and the locations of any files referenced by it) according to the following rules (source)[https://man.archlinux.org/man/extra/fontconfig/FcConfigGetFilename.3.en]:
    - If they start with `~`, resolve them relative to `$FONTCONFIG_SYSROOT/$HOME`
    - If they start with `/`, they are an absolute path relative to `$FONTCONFIG_SYSROOT` (unless they are pre-prefixed with `$FONTCONFIG_SYSROOT`, in which case it should not be prepended again)
    - If the do not start with `/` or `~`, they are relative to the system conf directories, which are defined by `$FONTCONFIG_SYSROOT` + each element of `$FONTCONFIG_PATH` (which is a path list). If `$FONTCONFIG_PATH` is undefined, you use a system-specific path that in practice seems like it's `/etc/fonts`

It's a mess, basically. I'll try to implement this.

whereswaldon:
I've pushed an update to implement the above. @benoitkugler I had to change some test data because it was including config files via a relative path like `<include>other.conf</include>`, but fontconfig doesn't seem to support that. `<dir>` can supply `prefix=relative` to get that behavior, but not `<include>`. Please sanity check me here. I couldn't find any instances of relative include paths in real fontconfig files, and I couldn't find anything in the docs to lead me to think it's allowed.

benoitkugler:
> I've pushed an update to implement the above. @benoitkugler I had to change some test data because it was including config files via a relative path like `<include>other.conf</include>`, but fontconfig doesn't seem to support that. `<dir>` can supply `prefix=relative` to get that behavior, but not `<include>`. Please sanity check me here. I couldn't find any instances of relative include paths in real fontconfig files, and I couldn't find anything in the docs to lead me to think it's allowed.

You're right, I've implemented that as it felt right, but I had not studied FcGetFilename close enough..
Thank you for your commit, which should be more accurate and close to real world configurations !

whereswaldon:
@benoitkugler I'll see about making the tests pass on windows next.

As for the one helvetica test that fails on my machine, I guess maybe we shouldn't expect the host system to have an arabic font that matches that family installed?

benoitkugler:
- > I think it's the user's fault if they request an impossible situation like a monospaced times font. Monospace seems (to me) to be a more specific request than the family. If a layout is requiring it, not respecting that seems likelier to break something than not respecting the family. My intuition is that if we can't find a family match that is monospaced, we should just return an arbitrary monospaced font.

Alright. Could we only support 'monospace' query and not 'monospace and this family' ? If so, this could be implemented by giving the 'monospace' family string the special behavior you specified : 

- font would need to pass the IsMonospace test
- if after family substitutions, no font is selected, we default to the first monospace font we have (outside the families)

The `Query` struct would not change : to query a monospace font, you pass `Query{Families: []string{"monospace"}}`

What do you think ?

whereswaldon:
> * > I think it's the user's fault if they request an impossible situation like a monospaced times font. Monospace seems (to me) to be a more specific request than the family. If a layout is requiring it, not respecting that seems likelier to break something than not respecting the family. My intuition is that if we can't find a family match that is monospaced, we should just return an arbitrary monospaced font.
> 
> 
> Alright. Could we only support 'monospace' query and not 'monospace and this family' ? If so, this could be implemented by giving the 'monospace' family string the special behavior you specified :
> 
>     * font would need to pass the IsMonospace test
> 
>     * if after family substitutions, no font is selected, we default to the first monospace font we have (outside the families)
> 
> 
> The `Query` struct would not change : to query a monospace font, you pass `Query{Families: []string{"monospace"}}`
> 
> What do you think ?

I'm not sure that that's viable. Isn't it fairly normal to use a typeface like Noto and want to use the monospaced version of it when embedding preformatted text? I'd hate for our logic to choose a different monospaced font. Or would it work as expected if you supplied `Query{Families:[]string{"noto", "monospace"}}`? I would expect it to return *either* a noto or a monospaced font.

fontscan/serialize.go
```diff
@@ -187,10 +187,17 @@ func (ff *fileFootprints) deserializeFrom(src []byte) error {
...
+		return i
+	}
+	return j
+}
+
 // serialize into binary format, compressed with gzip
 func (index systemFontsIndex) serializeTo(w io.Writer) error {
 	// version as uint16 + len as uint32 + somewhat the minimum size for a footprint
-	buffer := make([]byte, 6, 4+len(index)*(aspectSize+1+2))
+	buffer := make([]byte, 6, max(6, 4+len(index)*(aspectSize+1+2)))
```
benoitkugler, line 200: Good catch, thank you !

whereswaldon:
Okay, all tests pass on all CI platforms. :tada: 

As I mentioned in https://github.com/go-text/render/pull/8#discussion_r1232520950, I would like to alter this code to accept a user-provided logger instead of logging to the default logger unconditionally, but I figured that's a less important concern that getting the code and API correct.

benoitkugler:
> I'll own that there are relatively few _desktop_ systems with no arabic font at all, but imagine the case of running tests for a server-side project that wants to do text shaping. The environment may be a bare-minimum container, in which case there may not be any fonts. In general, I'm not sure that it's a good idea to write tests that rely upon the system to have any particular set of fonts installed, or at least not to run those by default. I still can't figure out why this test fails for me, since I actually do have arabic fonts.

Alright, fair enough !

opentype/testutils/utils.go
```diff
@@ -49,7 +49,9 @@ func FilenamesFS(t testing.TB, fs *embed.FS, dir string) []string {
 		if entry.IsDir() {
 			continue
 		}
-		filename := filepath.Join(dir, entry.Name())
+		// We should not use filepath.Join here because embed.FS still uses
```
benoitkugler, line 52: Oh ! Thanks, I've never noticed that..

benoitkugler:
> I'm not sure that that's viable. Isn't it fairly normal to use a typeface like Noto and want to use the monospaced version of it when embedding preformatted text? I'd hate for our logic to choose a different monospaced font. Or would it work as expected if you supplied `Query{Families:[]string{"noto", "monospace"}}`? I would expect it to return _either_ a noto or a monospaced font.

You are correct about `Query{Families:[]string{"noto", "monospace"}}`
But, huh, it seems that mixing family and monospace will be more complicated...
Also note that (as far as I now), monospace font usually include "mono" in their family name. So that specifying "noto" and "monospace" would not match "noto mono" (at least with the current implementation) ..

I'll need to think more about it. I'm a bit worried that it would complicate the semantics and also go far from other libraries behavior, but well...

whereswaldon:
> > I'm not sure that that's viable. Isn't it fairly normal to use a typeface like Noto and want to use the monospaced version of it when embedding preformatted text?
> 
> Well actually, the CSS spec does not provide a way for it. If you want Noto monospace, you have to explicitly ask for "NotoMono"

A bizarre oversight. Huh. That's a good datapoint, though it's also an opportunity for us to do better than CSS (if there is a reasonable way to do it).

> > I'd hate for our logic to choose a different monospaced font. Or would it work as expected if you supplied `Query{Families:[]string{"noto", "monospace"}}`? I would expect it to return _either_ a noto or a monospaced font.
> 
> You are correct about `Query{Families:[]string{"noto", "monospace"}}` But, huh, it seems that mixing family and monospace will be more complicated... Also note that (as far as I now), monospace font usually include "mono" in their family name. So that specifying "noto" and "monospace" would not match "noto mono" (at least with the current implementation) ..

Hmm. I'd really like for users to be able to say "I want a monospace font" and always get one. I'd also really like for users to be able to say "I want a Noto font" and "I want a monospace font" and end up with Noto Mono. However, it seems like that might be unrealistic given your discussion above?

Assuming we could add "IsMonospace" to the footprint, would it then be possible to do the following:

- Add `IsMonospace` as a `Query` field
- If the query has `IsMonospace` set to true:
  - When matching against footprints, only consider footprints with `IsMonospace`
  - If no footprints pass, repeat the match without enforcing `IsMonospace`

In this way, it is possible to get a monospace font for a specific family.

> 
> I'll need to think more about it. I'm a bit worried that it would complicate the semantics and also go far from other libraries behavior, but well...

Could you elaborate on how other libraries differ? And do you think that the approach that they take is better, or simply more conventional?

benoitkugler:
> Assuming we could add "IsMonospace" to the footprint, would it then be possible to do the following:
> 
> * Add `IsMonospace` as a `Query` field
> * If the query has `IsMonospace` set to true:
>   
>   * When matching against footprints, only consider footprints with `IsMonospace`
>   * If no footprints pass, repeat the match without enforcing `IsMonospace`
> 
> In this way, it is possible to get a monospace font for a specific family.

Yes that seems possible, indeed 

> Could you elaborate on how other libraries differ? And do you think that the approach that they take is better, or simply more conventional?

Well I only had rust font-kit in mind, which implements CSS rules. I'm hoping go-text will implement a CSS compatible behavior, but your approach seems just fine !

whereswaldon:
@benoitkugler When you have time, please look at https://github.com/go-text/typesetting/pull/63/commits/31d8574748ff28f29b134817c14106d2ecd41686 and let me know if you have a better approach in mind. I opted to create a second method so that we didn't need to change the `shaping.FontMap` interface. There are many other ways we could go about offering the metadata though, and I'd be happy to do something else if you have a preference.

whereswaldon:
It seems we have work to do for WASM graceful fallback. Or, at least, I do:

image

whereswaldon:
image

It was all on Gio's end. WASM currently doesn't provide any system fonts, but I've made Gio tolerant of having zero faces available so that we can at least present the non-text UI if the user didn't supply any faces *and* the system doesn't supply any.

benoitkugler:
> @benoitkugler When you have time, please look at [31d8574](https://github.com/go-text/typesetting/commit/31d8574748ff28f29b134817c14106d2ecd41686) and let me know if you have a better approach in mind. I opted to create a second method so that we didn't need to change the `shaping.FontMap` interface. There are many other ways we could go about offering the metadata though, and I'd be happy to do something else if you have a preference.

My preference would go towards actually updating the shaping.FontMap interface.
The current version was rather a draft, and I don't think it is really used in user code. 
I'm not even sure we need the interface at all, do we ?

That would add a disparity with the more elementary ResolveFace but since the two functions tackle different issues, it does not bother me that much. 

What are you thoughs ?

whereswaldon:
> > @benoitkugler When you have time, please look at [31d8574](https://github.com/go-text/typesetting/commit/31d8574748ff28f29b134817c14106d2ecd41686) and let me know if you have a better approach in mind. I opted to create a second method so that we didn't need to change the `shaping.FontMap` interface. There are many other ways we could go about offering the metadata though, and I'd be happy to do something else if you have a preference.
> 
> My preference would go towards actually updating the shaping.FontMap interface. The current version was rather a draft, and I don't think it is really used in user code. I'm not even sure we need the interface at all, do we ?
> 

> What are you thoughs ?

My reasoning was that the actual functionality of `ResolveFace` when used by the shaping package doesn't really need the metadata, so it places a burden on implementers of `shaping.FontMap` for no reason. As for whether we need it at all... I think it's a good idea to keep it as an interface. I can imagine people wanting to use go-text, but to plug in the system font provider library for the process of choosing faces. 

> That would add a disparity with the more elementary ResolveFace but since the two functions tackle different issues, it does not bother me that much.

I feel bad about this knowing how hard you worked on the simpler `ResolveFace`, but I don't currently have a use-case for it. Gio is using the `FontMap` approach and just paying the application startup cost of 200ms or so. The limitations of `ResolveFace` mean that it can't be used for fallback, which is one of the primary Gio use-cases for system fonts. Do *you* have a use-case for it? If not, it's possible that we don't need it at all.

benoitkugler:
> I feel bad about this knowing how hard you worked on the simpler `ResolveFace`, but I don't currently have a use-case for it.

No worries :)  Besides, I think we should back-port its logic into `Fontmap.ResolveFace`, since the decision to look for a new face or not is still relevant in this context.

>  Do you have a use-case for it? If not, it's possible that we don't need it at all.

I don't think so. I'll want to use the full featured `Fontmap`. The simpler `ResolveFace` was more a kind of prototype. 

> I can imagine people wanting to use go-text, but to plug in the system font provider library for the process of choosing faces.

Sure thing, but for now we never use a `Fontmap` as input of one of our functions...

> Gio is using the FontMap approach and just paying the application startup cost of 200ms or so. 

Is this new ? I thought Gio was using the simpler `ResolveFace` with a fixed slice of fonts setup by the developper.

whereswaldon:
> > I feel bad about this knowing how hard you worked on the simpler `ResolveFace`, but I don't currently have a use-case for it.
> 
> No worries :) Besides, I think we should back-port its logic into `Fontmap.ResolveFace`, since the decision to look for a new face or not is still relevant in this context.
> 
> > Do you have a use-case for it? If not, it's possible that we don't need it at all.
> 
> I don't think so. I'll want to use the full featured `Fontmap`. The simpler `ResolveFace` was more a kind of prototype.
> 
> > I can imagine people wanting to use go-text, but to plug in the system font provider library for the process of choosing faces.
> 
> Sure thing, but for now we never use a `Fontmap` as input of one of our functions...

We use on [here](https://pkg.go.dev/github.com/go-text/typesetting/shaping#SplitByFace), which is the functionality Gio is consuming to load system fonts.

> > Gio is using the FontMap approach and just paying the application startup cost of 200ms or so.
> 
> Is this new ? I thought Gio was using the simpler `ResolveFace` with a fixed slice of fonts setup by the developper.

This is the approach that I'm prototyping, yeah. `ResolveFace` doesn't let us resolve by rune coverage, and we're constantly running into "My application keeps showing tofu when a user inputs text in language X" problems. It's clear that we need to be able to fall back to the system-installed fonts to display unexpected multilingual text, and only the fontmap gives us the ability to do that.

There's still a fixed slice of fonts pre-loaded by the developer, but now we use the FontMap to load additional system fonts as-needed.

whereswaldon:
I've got system font loading working on Android, but it required some tweaks (the commits that I've just pushed).

image

The most frustrating problem was that I couldn't find a safe directory to use to store our cache automatically, so I've officially pushed that responsibility onto the calling application. In theory the caller is a proper Android app with the ability to use JNI to query the proper cache directory for us.

whereswaldon:
Screenshot 2023-06-20 at 3 52 18 PM

Things are looking good on macOS also, though I think we knew that already.

whereswaldon:
Okay, I was able to get iOS working as well. I had to update the code to not assume the cache directory already exists, but that makes sense.

image

whereswaldon:
Windows just works without a hitch:

image

benoitkugler:
> We use one [here](https://pkg.go.dev/github.com/go-text/typesetting/shaping#SplitByFace), which is the functionality Gio is consuming to load system fonts.
> 
> > >

Wow, my mistake, I forgot we have also implemented the general approach.

OK, so it is probably best to keep your approach (not changing the `Fontmap` interface) and bear with the "duplicated" `ResolveFace` and `ResolveFaceAndMetadata`.

whereswaldon:
> > We use one [here](https://pkg.go.dev/github.com/go-text/typesetting/shaping#SplitByFace), which is the functionality Gio is consuming to load system fonts.
> > > >
> 
> Wow, my mistake, I forgot we have also implemented the general approach.

No worries!

> OK, so it is probably best to keep your approach (not changing the `Fontmap` interface) and bear with the "duplicated" `ResolveFace` and `ResolveFaceAndMetadata`.

I don't mind changing the API if we have a technical reason to, but the current thing is certainly the least churn. Is there a structure that seems superior?

benoitkugler:
> > > We use one [here](https://pkg.go.dev/github.com/go-text/typesetting/shaping#SplitByFace), which is the functionality Gio is consuming to load system fonts.
> > > > >
> > 
> > 
> > Wow, my mistake, I forgot we have also implemented the general approach.
> 
> No worries!
> 
> > OK, so it is probably best to keep your approach (not changing the `Fontmap` interface) and bear with the "duplicated" `ResolveFace` and `ResolveFaceAndMetadata`.
> 
> I don't mind changing the API if we have a technical reason to, but the current thing is certainly the least churn. Is there a structure that seems superior?

We could perhaps use a new method `Fontmap.Metadata()` similar to `FaceLocation`. Cleaner API, but probably a bit less performant when the number of faces actually used grows

whereswaldon:
I was unable to set up a FreeBSD system capable of testing this within a few hours, so I think we're just going to have to hope it works. I did get confirmation from a friend that it's working on OpenBSD.

whereswaldon:
> > > > We use one [here](https://pkg.go.dev/github.com/go-text/typesetting/shaping#SplitByFace), which is the functionality Gio is consuming to load system fonts.
> > > > > >
> > > 
> > > 
> > > Wow, my mistake, I forgot we have also implemented the general approach.
> > 
> > 
> > No worries!
> > > OK, so it is probably best to keep your approach (not changing the `Fontmap` interface) and bear with the "duplicated" `ResolveFace` and `ResolveFaceAndMetadata`.
> > 
> > 
> > I don't mind changing the API if we have a technical reason to, but the current thing is certainly the least churn. Is there a structure that seems superior?
> 
> We could perhaps use a new method `Fontmap.Metadata()` similar to `FaceLocation()`. Cleaner API, but probably a bit less performant when the number of faces actually used grows

I wouldn't want to traverse the entire set of loaded fonts in the way that `FaceLocation` does. I'd be open to this if we used a more efficient strategy. It seems like that would require maintaining multiple maps of the loaded fonts though (in order to support querying by different keys), which introduces many opportunities for mistakes. Would you prefer that?

benoitkugler:
> I wouldn't want to traverse the entire set of loaded fonts in the way that `FaceLocation` does. I'd be open to this if we used a more efficient strategy. It seems like that would require maintaining multiple maps of the loaded fonts though (in order to support querying by different keys), which introduces many opportunities for mistakes. Would you prefer that?

Hum.. a map[Face] -> (Location, Metadata) would be sufficient for efficient `FaceLocation` and `FaceMetadata` methods right ?
Maybe it is not to dangerous to use this approach : I see only two writes to the `cache` field, and we could use an helper method to ensure consistency.

whereswaldon:
@benoitkugler I've tried to implement a separate `Metadata()` method. Let me know what you think.

As far as I currently know, the only TODOs on this branch are:

- [ ] rebase on `main` to resolve the harfbuzz-related conflicts
- [ ] implement something for resolving monospace fonts

I'd appreciate it if you could handle the rebase, and I believe that you've already got some work for monospace stuff. Is there anything I can do to help right now? If not, I'll focus on the Gio side of this feature.

whereswaldon:
I found a hole in the API that I've tried to plug in #80. Since it's new API design and a feature we haven't already discussed, I thought it made sense to discuss it in a separate draft PR with this as the base branch.

benoitkugler:
> @benoitkugler I've tried to implement a separate `Metadata()` method. Let me know what you think.
> 
> As far as I currently know, the only TODOs on this branch are:
> 
> * [ ]  rebase on `main` to resolve the harfbuzz-related conflicts
> * [ ]  implement something for resolving monospace fonts
> 
> I'd appreciate it if you could handle the rebase, and I believe that you've already got some work for monospace stuff. Is there anything I can do to help right now? If not, I'll focus on the Gio side of this feature.

Thank you very much for the many improvements ! There were so many gaps you have filled ! 
I can indeed get a stab at the two remaining points.

whereswaldon:
@benoitkugler thinking on it (and discussing the issue with @dominikh), I'm starting to think I was wrong to want a special case for monospace fonts. It's expensive for us to check if a font is monospace, and no other font resolution platform implements it like this. They expect you to either specify a known monospace font by family name or to rely on the generic family names like `monospace`. Maybe that is the better option.

Based on your current view of things, does it seem better to not handle that as a special case, or to persist in adding special resolution mechanisms for it?

whereswaldon:
I think we may need to do some extra work to honor [this CSS requirement](https://www.w3.org/TR/css-fonts-3/#font-family-desc)

> If the font family name [in @font-face] is the same as a font family available in a given user's environment, it effectively hides the underlying font for documents that use the stylesheet. This permits a web author to freely choose font-family names without worrying about conflicts with font family names present in a given user's environment.

Right now, manually loading a font into the fontmap doesn't prevent a system font of the same name from being used. In fact, the system font may win always depending on load ordering.

I'm really struggling with how to best ensure that fonts loaded by applications are used whenever possible. #80 handles the case in which there are no system fonts loaded, allowing us to fall back to manually loaded fonts. However, in the presence of system fonts, the default rule will add `sans-serif` and derivatives, ensuring that a manually loaded sans-serif font is never chosen if the system provides any.

Maybe that's okay, as we expect application authors to request their custom faces by name. It's simply irritating because Gio historically didn't work this way, and it's unclear how to migrate to the new behavior smoothly.

benoitkugler:
> @benoitkugler thinking on it (and discussing the issue with @dominikh), I'm starting to think I was wrong to want a special case for monospace fonts. It's expensive for us to check if a font is monospace, and no other font resolution platform implements it like this. They expect you to either specify a known monospace font by family name or to rely on the generic family names like `monospace`. Maybe that is the better option.
> 
> Based on your current view of things, does it seem better to not handle that as a special case, or to persist in adding special resolution mechanisms for it?

Ah this is very interesting !

As far as my personal use of go-text is concerned, I would not have use the `IsMonospace` Query flag, so...

I've noticed that the flag behavior may be surprising : for instance, for `Query{Families: []string{"NotoSans"}, EnforceMonospace: true}`, the NotoSansMono would not always match ! (because the family names do no match, and assuming they are other monospace fonts in the database).

So the only "well-defined" use case (in my opinion), is to combine a generic family like "serif" or "sans-serif" with the IsMonospace flag. Does it often happens in practice ?

benoitkugler:
> I think we may need to do some extra work to honor [this CSS requirement](https://www.w3.org/TR/css-fonts-3/#font-family-desc)
> 
> > If the font family name [in @font-face] is the same as a font family available in a given user's environment, it effectively hides the underlying font for documents that use the stylesheet. This permits a web author to freely choose font-family names without worrying about conflicts with font family names present in a given user's environment.
> 
> Right now, manually loading a font into the fontmap doesn't prevent a system font of the same name from being used. In fact, the system font may win always depending on load ordering.
> 
> I'm really struggling with how to best ensure that fonts loaded by applications are used whenever possible. #80 handles the case in which there are no system fonts loaded, allowing us to fall back to manually loaded fonts. However, in the presence of system fonts, the default rule will add `sans-serif` and derivatives, ensuring that a manually loaded sans-serif font is never chosen if the system provides any.
> 
> Maybe that's okay, as we expect application authors to request their custom faces by name. It's simply irritating because Gio historically didn't work this way, and it's unclear how to migrate to the new behavior smoothly.

Could this be handled by a tag on the footprint ("user"/"system"), then giving higher priority to "user" ? The fonts with the same family name would have the same score, so this additional sort would yield the expected order (I think !)

whereswaldon:
> > @benoitkugler thinking on it (and discussing the issue with @dominikh), I'm starting to think I was wrong to want a special case for monospace fonts. It's expensive for us to check if a font is monospace, and no other font resolution platform implements it like this. They expect you to either specify a known monospace font by family name or to rely on the generic family names like `monospace`. Maybe that is the better option.
> > Based on your current view of things, does it seem better to not handle that as a special case, or to persist in adding special resolution mechanisms for it?
> 
> Ah this is very interesting !
> 
> As far as my personal use of go-text is concerned, I would not have use the `IsMonospace` Query flag, so...
> 
> I've noticed that the flag behavior may be surprising : for instance, for `Query{Families: []string{"NotoSans"}, EnforceMonospace: true}`, the NotoSansMono would not always match ! (because the family names do no match, and assuming they are other monospace fonts in the database).
> 
> So the only "well-defined" use case (in my opinion), is to combine a generic family like "serif" or "sans-serif" with the IsMonospace flag. Does it often happens in practice ?

In practice, I think Gio users are spoiled by our previous font matcher explicitly allowing you to say "give me the first available monospace font", but we shouldn't let that drive the design.

Given that the flag behaves in a counterintuitive way unless you use it with generic families, I think we should drop it.

We may even be able to drop the monospace information from footprint and metadata. It was all an effort to implement this older Gio behavior that I now think was both poorly-defined and cannot be implemented efficiently for system fonts. Of course, it cay stay if someone does have a use for it.

I feel quite silly, if I'm honest. I approached this with the goal of implementing a compatible interface to what Gio already did, but Gio was doing something super unrealistic. The more that I come to understand the mechanics here, the more I think Gio has to change to match the reality of what metadata is available.

whereswaldon:
> > I think we may need to do some extra work to honor [this CSS requirement](https://www.w3.org/TR/css-fonts-3/#font-family-desc)
> > > If the font family name [in @font-face] is the same as a font family available in a given user's environment, it effectively hides the underlying font for documents that use the stylesheet. This permits a web author to freely choose font-family names without worrying about conflicts with font family names present in a given user's environment.
> > 
> > 
> > Right now, manually loading a font into the fontmap doesn't prevent a system font of the same name from being used. In fact, the system font may win always depending on load ordering.
> > I'm really struggling with how to best ensure that fonts loaded by applications are used whenever possible. #80 handles the case in which there are no system fonts loaded, allowing us to fall back to manually loaded fonts. However, in the presence of system fonts, the default rule will add `sans-serif` and derivatives, ensuring that a manually loaded sans-serif font is never chosen if the system provides any.
> > Maybe that's okay, as we expect application authors to request their custom faces by name. It's simply irritating because Gio historically didn't work this way, and it's unclear how to migrate to the new behavior smoothly.
> 
> Could this be handled by a tag on the footprint ("user"/"system"), then giving higher priority to "user" ? The fonts with the same family name would have the same score, so this additional sort would yield the expected order (I think !)

Yes, I definitely think this would work.

benoitkugler:
> I feel quite silly, if I'm honest.

Thank you for you intellectual honesty, very much appreciated !

> Given that the flag behaves in a counterintuitive way unless you use it with generic families, I think we should drop it.
> 
> We may even be able to drop the monospace information from footprint and metadata. It was all an effort to implement this older Gio behavior that I now think was both poorly-defined and cannot be implemented efficiently for system fonts. Of course, it cay stay if someone does have a use for it.

Alright, let's drop it then. 

I think we can keep it in the metadata package, but ensure we don't pay the price to compute it in fontscan. I'll try to do so tomorrow.


> > Could this be handled by a tag on the footprint ("user"/"system"), then giving higher priority to "user" ? The fonts with the same family name would have the same score, so this additional sort would yield the expected order (I think !)
> 
> Yes, I definitely think this would work.

Great, I'll try to tackle it as well tomorrow.

benoitkugler:
By the way, should we keep the `FindFont` API (filename based, without rune coverage) ?
I've initialy written it to propose a faster alternative to the `Fontmap` API, but do you think it is still useful for the toolkits ? (It won't be for my use case.)

whereswaldon:
> By the way, should we keep the `FindFont` API (filename based, without rune coverage) ? I've initialy written it to propose a faster alternative to the `Fontmap` API, but do you think it is still useful for the toolkits ? (It won't be for my use case.)

I don't think Gio needs it, and I suspect Fyne will be in the same boat. If we want to resolve a system font, we're probably trying to fill a gap in rune coverage. We might just be loading it for the sake of having a prettier font, but the scan isn't actually as slow as I would have guessed, so it's usable in practice by the toolkits anyway.

benoitkugler:
I've applied the changes we have discussed, along with some memory allocation optimization.

I would like to go a bit further, but it will require a small API change in `metadata`, so I'll open a separate PR.

fontscan/substitutions.go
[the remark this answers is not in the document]
andydotxyz: Should it be `langEqualsAndNotFamily` instead then? "no" implies empty whereas "not" is a negation which seems to be what you describe.

andydotxyz:
Sorry all, just catching up on this long PR - great work, and highly relevant to a task I am about to pick up too. Apologies for the delay.

> * [ ]  ~FreeBSD~ I was unable to get a working graphical FreeBSD system in a reasonable amount of time, so I think we're just going to have to hope it works similarly enough to OpenBSD for now.

Can I help on this one? Fyne has a FreeBSD test target which I should be able to run appropriate test / searches on.

> I feel quite silly, if I'm honest. I approached this with the goal of implementing a compatible interface to what Gio already did, but Gio was doing something super unrealistic. The more that I come to understand the mechanics here, the more I think Gio has to change to match the reality of what metadata is available.

I think you're being harsh on yourself here. Fyne also expects the ability to have a "monospace font" automatically known - as you say it is a user expectation that typing code blocks etc will pick appropriately.
I don't know if a query lookup in the way discussed is necessarily the only way to do it - but as the main text/font library for Go graphics I think we do need a solution to this problem.
Perhaps, as we have platform specific code, we can have a simplified version of "what is the normal monospace font" - I.e. not a full scan, but something we know exists in the OS install?

If it's not possible to do something like that I think Fyne, like Gio, will have to continue packaging a monospace font, or implementing complex lookup logic to determine what is a good monospace guess and it feels like that belongs better here than in the toolkit.

benoitkugler:
> I don't know if a query lookup in the way discussed is necessarily the only way to do it - but as the main text/font library for Go graphics I think we do need a solution to this problem. Perhaps, as we have platform specific code, we can have a simplified version of "what is the normal monospace font" - I.e. not a full scan, but something we know exists in the OS install?
> 
> If it's not possible to do something like that I think Fyne, like Gio, will have to continue packaging a monospace font, or implementing complex lookup logic to determine what is a good monospace guess and it feels like that belongs better here than in the toolkit.

Note that, with the current solution, using `Query{Families: []string{"monospace"}}` is a valid way to get a monospace font (providing the system has one), which hopefully is enough in practice.
The only behavior not covered is "I want a _serif_ font which is mono-spaced", but I hope it is not required too often..

benoitkugler:
> > * [ ]   ~FreeBSD~ I was unable to get a working graphical FreeBSD system in a reasonable amount of time, so I think we're just going to have to hope it works similarly enough to OpenBSD for now.
> 
> Can I help on this one? Fyne has a FreeBSD test target which I should be able to run appropriate test / searches on.

That would be great ! I guess that checking if the tests pass are a first start, and then trying to actually display something using the `Fontmap` would be nice. 
A pseudo code you could use : 
```
fm := fontscan.NewFontMap()
fm.UseSystemFonts() 
input := shaping.Input{Text: []rune("my text")}
inputs := shaping.SplitByFace(input, fm)
// then shape with harfbuzz and display outputs
```
Perhaps @whereswaldon has other suggestions ?

whereswaldon:
> > I feel quite silly, if I'm honest. I approached this with the goal of implementing a compatible interface to what Gio already did, but Gio was doing something super unrealistic. The more that I come to understand the mechanics here, the more I think Gio has to change to match the reality of what metadata is available.
> 
> I think you're being harsh on yourself here. Fyne also expects the ability to have a "monospace font" automatically known - as you say it is a user expectation that typing code blocks etc will pick appropriately. I don't know if a query lookup in the way discussed is necessarily the only way to do it - but as the main text/font library for Go graphics I think we do need a solution to this problem. Perhaps, as we have platform specific code, we can have a simplified version of "what is the normal monospace font" - I.e. not a full scan, but something we know exists in the OS install?
> 
> If it's not possible to do something like that I think Fyne, like Gio, will have to continue packaging a monospace font, or implementing complex lookup logic to determine what is a good monospace guess and it feels like that belongs better here than in the toolkit.

The ability to resolve a monospace font is easy enough. As @benoitkugler says, you just say `monospace` as the family. What Gio used to offer was a way to specify all of:

- family name
- style (italic/normal)
- weight
- variation (monospace/smallcaps)

However, in practice the variation is part of the family. Those variations aren't exposed in font metadata that can be queried efficiently. Allowing callers to request "Go" as the family and "Mono" as the variation seems fine on the surface, but that can't be queried efficiently. However, such users can just request "Go Mono" as the family to get what they wanted. Gio dodged this by having the user annotate the fonts manually to specify their variation, but we can't do that for system fonts.

I'm planning to just drop the "variation" from the API, expecting users to encode what they want in the family name instead. This matches how every other font resolution system seems to work. I'll also be adding a way to specify the fallback chain so that you can say `"Go Mono, Courier, monospace"`. Gio currently doesn't allow this sort of configurable fallback.

whereswaldon:
> > > * [ ]    ~FreeBSD~ I was unable to get a working graphical FreeBSD system in a reasonable amount of time, so I think we're just going to have to hope it works similarly enough to OpenBSD for now.
> > 
> > 
> > Can I help on this one? Fyne has a FreeBSD test target which I should be able to run appropriate test / searches on.
> 
> That would be great ! I guess that checking if the tests pass are a first start, and then trying to actually display something using the `Fontmap` would be nice. A pseudo code you could use :
> 
> ```
> fm := fontscan.NewFontMap()
> fm.UseSystemFonts() 
> input := shaping.Input{Text: []rune("my text")}
> inputs := shaping.SplitByFace(input, fm)
> // then shape with harfbuzz and display outputs
> ```
> 
> Perhaps @whereswaldon has other suggestions ?

@andydotxyz Here's a complete program that you can use to test font support:

```go
package main

import (
	"log"
	"os"

	"github.com/go-text/typesetting/fontscan"
)

func main() {
	cwd, err := os.Getwd()
	if err != nil {
		panic(err)
	}
	fontMap := fontscan.NewFontMap(log.Default())
	err = fontMap.UseSystemFonts(cwd)
	if err != nil {
		panic(err)
	}
	fontMap.SetQuery(fontscan.Query{
		Families: []string{""},
	})
	face := fontMap.ResolveFace('a')
	family, aspect := fontMap.FontMetadata(face.Font)
	log.Printf("got face: %p %s %#+v", face, family, aspect)
}
```

If the final log line prints a real face family name and aspect, we successfully resolved a system font. I don't think displaying text graphically is strictly necessary, though it was how I was doing my testing.

whereswaldon:
From where I stand, this PR is ready for review. I'm not aware of feature gaps or fundamental concerns with the approach. Since clicking the "ready for review" button doesn't do anything except legitimize the PR in GitHub, I think I'll do that now. We can continue with code review as per any other PR.

whereswaldon:
A friend confirmed that it's working on FreeBSD for me, so all of my target platforms are covered.

fontscan/fontmap.go
```diff
@@ -0,0 +1,489 @@
...
+//	// at least one of the following calls
+//	fontMap.UseSystemFonts() // error handling omitted
+//	fontMap.AddFont(font1, "font1") // error handling omitted
+//	fontMap.AddFace(alreadyParsed, metadata) // error handling omitted
+//
+//	// set the font description
+//	fontMap.SetQuery(Query{Families: []string{"Arial", "serif"}}) // regular Aspect
+//
+//	// `fontMap` is now ready for text shaping, using the `ResolveFace` method
+//
```
whereswaldon, line 43: We should convert this into a proper example. The API has already drifted so that this isn't valid at least once.

fontscan/fontmap.go
```diff
@@ -0,0 +1,489 @@
...
+	}
+}
+
+// UseSystemFonts loads the system fonts and adds them to the font map.
+// This method is safe for concurrent use, but should only be called once
+// per font map.
+// The first call of this method trigger a rather long scan.
+// A per-application on-disk cache is used to speed up subsequent initialisations.
+// Callers can provide an appropriate directory path within which this cache may be
+// stored.
```
whereswaldon, line 97: Need to update this to say what happens if the caller doesn't provide a path.

fontscan/readme.md
```diff
@@ -0,0 +1,76 @@
...
+To support apps displaying a large number of scripts, a GUI toolkit want to leverage
+the fonts provided by the OS (system fonts). This ensure correct rendering, and alleviate
+the work required by the developper.
+
+### Markup language renderers
+
+The second use case deals with the rendering of markup documents (think HTML or SVG), where the author provide _hints_ about the fonts that should be used. In this case, the two main requirements are
+
+- use the whole system fonts to find the best match with respect to the author intention
+- handle a large rune coverage : many scripts may be present in the same document, so that we can't specify at build time the set of fonts we will use
```
whereswaldon, line 21: These use-cases aren't all that different anymore, so we can probably merge these sections.

fontscan/readme.md
```diff
@@ -0,0 +1,76 @@
...
+- for the Arial input family, [Arial, Arimo, sans-serif, DejaVu Sans, Verdana] would be matched
+
+To respect the user request, the order of the list is significant (first entries have higher priority).
+
+`FontMap.SetQuery` apply a list of hard-coded subsitutions, extracted from
+Fontconfig configurations files.
+
+### Style matching
+
+`FontMap.SetQuery` takes an optionnal argument describing the style of
```
whereswaldon, line 61: `s/optionnal/optional/`

fontscan/scandir_old.go
```diff
@@ -0,0 +1,132 @@
...
+		return err
+	}
+
+	err := filepath.Walk(dir, walkFn)
+
+	return err
+}
+
+// DirEntry is a copy of the Go 1.16+ fs.DirEntry interface.
+type DirEntry interface {
```
whereswaldon, line 52: I think we should be able to unexport this.

fontscan/substitutions.go
whereswaldon: I've created #82 to remind us about this, so I'm going to resolve this conversation.

whereswaldon (commented):
Just took a pass through and made some comments for things that need updating. I'll handle them all.

benoitkugler (commented):
Thank you for the cleanups ! I'm fine with merging this iteration. 

And thank you all for the great work on this really non trivial task !!

andydotxyz:
My apologies for delays once again.
FreeBSD gets this from the test app: `got face: 0x4000126810 luxisans metadata.Aspect{Style:0x1, Weight:400, Stretch:1}`
And latest Darwin works too `got face: 0x14003c7ba40 verdana metadata.Aspect{Style:0x1, Weight:400, Stretch:1}`

Were there iOS and Android tests completed as well or was that out of scope at this time?

andydotxyz:
iOS results are positive. but Android seems to crash, investigating.

Screenshot 2023-07-04 at 16 24 07

andydotxyz:
Android log...

```
07-04 16:25:49.404  5512  5512 I Fyne    : android device reports unknown density: 440
07-04 16:25:49.406  5512  5546 I Fyne    : using system font dirs ["/data/fonts" "/system/fonts"]
07-04 16:25:49.406  5512  5546 I Fyne    : error walking font directory "/data/fonts": open /data/fonts: permission denied
07-04 16:25:49.531  5512     0 E Go      : panic: updating cache: open /font_index.cache: read-only file system
07-04 16:25:49.531  5512     0 E Go      : 
07-04 16:25:49.531  5512     0 E Go      : goroutine 10 [running]:
07-04 16:25:49.531  5512     0 E Go      : main.main()
07-04 16:25:49.531  5512     0 E Go      :      /Users/andy/Code/Fyne/fyne/cmd/hello/main.go:22 +0x25c
07-04 16:25:49.531  5512     0 E Go      : fyne.io/fyne/v2/internal/driver/mobile/app/callfn.CallFn(0x77196fb3a0)
07-04 16:25:49.531  5512     0 E Go      :      /Users/andy/Code/Fyne/fyne/internal/driver/mobile/app/callfn/callfn_arm64.s:10 +0x20
07-04 16:25:49.531  5512     0 E Go      : created by fyne.io/fyne/v2/internal/driver/mobile/app.callMain
07-04 16:25:49.531  5512     0 E Go      :      /Users/andy/Code/Fyne/fyne/internal/driver/mobile/app/android.go:118 +0x1a0
```

Is it OK to ask for this panic to be fixed before merging?

andydotxyz:
Oh, my apologies I see now that it was expected that the demo code would have some android item in it.
Resolved once utilised appropriately:

Screenshot 2023-07-04 at 16 40 36

andydotxyz (approved):
Phenomenal effort, working on all platforms too - well done!

andydotxyz:
One last comment, not sure if it matters or not... But it may fail to find matching faces for some runes - emoji was a test that seemed to fail for me:

```
2023/07/04 16:47:01 using system font dirs ["/Library/Fonts" "/System/Library/Fonts" "/Users/andy/Library/Fonts"]
2023/07/04 16:47:01 No font matched for [] and rune U+1F600 (😀) -> returning arbitrary face
```

benoitkugler:
> One last comment, not sure if it matters or not... But it may fail to find matching faces for some runes - emoji was a test that seemed to fail for me:
> 
> ```
> 2023/07/04 16:47:01 using system font dirs ["/Library/Fonts" "/System/Library/Fonts" "/Users/andy/Library/Fonts"]
> 2023/07/04 16:47:01 No font matched for [] and rune U+1F600 (😀) -> returning arbitrary face
> ```
> 
> Could it be a multi-byte rune issue? Or are emoji fonts problematic

Very interesting point. One could expect the FontMap to be smart enough to select a proper "emoji font" to handle such runes, but it is not the case. At the FontMap level, the solution is to explicitly ask for an emoji font, by providing `Query{Families: []string{"emoji"}}`.
This will require more work from the toolkit. I think we may want to add in the future an "emoji iterator", whose job would be to detect emoji runes, so that the Fontmap consumers may update the query accordingly.

benoitkugler:
I've opened #83 to export the common generic families;  `Serif`, `SansSerif`, `Emoji` and `Monospace` being probably the most useful ones.

whereswaldon:
I've rebased this atop main, as GitHub was unable to resolve the conflicts itself.

whereswaldon:
My rebase made no logic changes (just had to fix conflicts in `go.mod` over and over and over and a couple of obvious other conflicts). As CI is still happy, I think this can be merged.

PR #65, opened 2023-04-26
Address possible panics in locale parsing

### LINKED ISSUE #64: Test fails in harfbuzz package when LC_ALL is not set
I have a zsh environment on my M1 Macbook Air running macOS Ventura 13.2.1
When I run the `locale` utility I get the following result:
```
ml) rene@Fresno harfbuzz % locale
LANG=""
LC_COLLATE="C"
LC_CTYPE="UTF-8"
LC_MESSAGES="C"
LC_MONETARY="C"
LC_NUMERIC="C"
LC_TIME="C"
LC_ALL=
```
Note that LC_ALL is empty and LC_CTYPE is UTF-8.
When running go test in the harfbuzz folder it crashes in the function NewOTTagsFromScriptAndLanguage

```ml) rene@Fresno harfbuzz % go test .
--- FAIL: TestEmojisSequences (0.00s)
panic: runtime error: index out of range [5] with length 5 [recovered]
	panic: runtime error: index out of range [5] with length 5

goroutine 7 [running]:
testing.tRunner.func1.2({0x1092c56e0, 0x1400001b260})
	/Users/rene/sdk/go1.20.3/src/testing/testing.go:1526 +0x1c8
testing.tRunner.func1()
	/Users/rene/sdk/go1.20.3/src/testing/testing.go:1529 +0x384
panic({0x1092c56e0, 0x1400001b260})
	/Users/rene/sdk/go1.20.3/src/runtime/panic.go:884 +0x204
github.com/go-text/typesetting/harfbuzz.NewOTTagsFromScriptAndLanguage(0x16f458?, {0x14000018425, 0x5})
	/Users/rene/Projects/tdewolff/typesetting/harfbuzz/ot_tag.go:318 +0x240
```
The string it is trying to analyze only has "UTF-8" in it and it therefore fails in the code here:

```go
	for s = 1; s < len(langStr); s++ { // s index in lang_str
		if langStr[s-1] == '-' && langStr[s+1] == '-' {
```
The langStr[s+1] accesses the character after the 8 (from the string "UTF-8").
A zero terminated C string would have a zero character. In Go that is an invalid access. 

If I set LC_ALL to en_US.UTF-8 and I run go test again the crash is gone and the test succeeds.  

Can this code be made more robust? 
I've spend several hours trying to figure out why a program I was using crashes.

andydotxyz:
Fixes #64

### DIFF
--- a/harfbuzz/ot_tag.go
+++ b/harfbuzz/ot_tag.go
@@ -314,8 +314,8 @@ func NewOTTagsFromScriptAndLanguage(script language.Script, language language.La
 			privateUseSubtag = langStr
 		} else {
 			var s int
-			for s = 1; s < len(langStr)-1; s++ { // s index in lang_str
-				if langStr[s-1] == '-' && langStr[s+1] == '-' {
+			for s = 1; s < len(langStr); s++ { // s index in lang_str
+				if s < len(langStr)-1 && langStr[s-1] == '-' && langStr[s+1] == '-' {
 					if langStr[s] == 'x' {
 						privateUseSubtag = langStr[s:]
 						if limit == -1 {
--- a/harfbuzz/ot_tag.go
+++ b/harfbuzz/ot_tag.go
@@ -310,11 +310,11 @@ func NewOTTagsFromScriptAndLanguage(script language.Script, language language.La
 		langStr := languageToString(language)
 		limit := -1
 		privateUseSubtag := ""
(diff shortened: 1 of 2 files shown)


whereswaldon (approved):
This does seem like it eliminates the possibility of panicking here. I wonder whether more robust approaches than string indexing might be viable, but we can explore that some other time.

benoitkugler (approved):
Thank you for the fix!

PR #66, opened 2023-04-27
Simplify harfbuzz language parsing

benoitkugler:
As a follow up of #65, this PR extracts and simplifies a bit the processing of language private tags.

Also, I've hidden the `newOTTagsFromScriptAndLanguage`, since it is mainly internal to the Harfbuzz implementation.
(Note that we may want to use this function to check if a particular feature is supported for a particular language/script pair, but this is not for today, and we would probably export a different API.)


whereswaldon (approved):
So we're using the new code generator output and exporting a new `Language.SplitExtensionTags`. Looks good to me. Thanks!

andydotxyz (approved):

PR #70, opened 2023-05-04
opentype/api/font: try to avoid extra allocations

egonelbre:
Low hanging fruit in some performance.

The out slice was reallocated several times, we can use the length of contour points as a quick guess.

```
                   │ before.txt~ │             after.txt~              │
                   │   sec/op    │   sec/op     vs base                │
  BuildSegments-32   2.435µ ± 4%   1.367µ ± 2%  -43.87% (p=0.000 n=10)

                   │ before.txt~  │              after.txt~              │
                   │     B/op     │     B/op      vs base                │
  BuildSegments-32   3.484Ki ± 0%   1.750Ki ± 0%  -49.78% (p=0.000 n=10)

                   │ before.txt~ │             after.txt~             │
                   │  allocs/op  │ allocs/op   vs base                │
  BuildSegments-32    7.000 ± 0%   1.000 ± 0%  -85.71% (p=0.000 n=10)
```

### DIFF
--- a/opentype/api/font/renderer.go
+++ b/opentype/api/font/renderer.go
@@ -150,11 +150,16 @@ func midPoint(p, q api.SegmentPoint) api.SegmentPoint {
 
 // build the segments from the resolved contour points
 func buildSegments(points []contourPoint) []api.Segment {
+	if len(points) == 0 {
+		return nil
+	}
+
 	var (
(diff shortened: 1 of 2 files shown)


benoitkugler (approved):
Looks good to me, thank you !

whereswaldon (approved):
Nice work! Thanks for picking some of that low-hanging fruit.

PR #68, opened 2023-05-03
Use an iterator-style API to consume shaped text in line wrapper

whereswaldon:
This PR is part of my ongoing effort to refactor the line wrapper with the goal of making it both easier to understand and able to support additional features.

We've long discussed wanting to be able to make text shaping an incremental process instead of having to do it all up front. Having and API that works in terms of iterators makes that much easier. As such, I've restructured the line wrapper to consume an iterator API that is expected to provide shaped text. I've only implemented an iterator backed by a pre-shaped slice for now, but it shouldn't be terribly hard to implement a lazy-shaping iterator with a compatible API.

However, introducing an iterator complicated some shaping operations and triggered some extra allocations (we couldn't just reach into the slice storage of the shaped text arbitrarily anymore), which caused a nontrivial performance decrease.

---

This section is obsolete, but left for reference; skip to the next horizontal rule.

To make up for this, I've introduced `WrapBuffer`, an optional buffer types that callers can provide to the line wrapper to allow it to avoid allocating. Everything will work as expected if callers provide the zero value, but providing a non-zero buffer will potentially eliminate allocations altogether if the wrapping process is simple enough.

It has a rather dramatic impact on performance across the board, in many places improving on the pre-iterator implementation's performance and memory use.

I know that this PR makes quite a few API changes, but:

- I think we *need* an iterable way to do this, so some kind of breaking change was coming eventually for that. If we have concerns about *this* iterator API, I'm happy to explore those.
- It seems like a major performance bottleneck for this algorithm is allocation, so I'd also argue that we need to provide a way to do that more efficiently. I've opted to make it a caller-provided thing so that we don't have to impose additional restrictions on how to use the wrapped text. We could instead make these buffers into fields on the wrapper and tell the caller "do not wrap more text until you have either copied or finished with the current batch". That seemed easier to mess up though. Happy to discuss this as well.

Below are my benchmarks:

- `before-iterator.txt` is from `main`
- `iterator.txt` is from the first commit in this PR
- `after-buffering.txt` shows the performance increase from adding the allocation buffer, as well as including benchmarks for the unbuffered case.

<details>
<summary>Benchmark results</summary>
<pre>
goos: linux
goarch: amd64
pkg: github.com/go-text/typesetting/shaping
cpu: Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
                                                 │ before-iterator.txt │             iterator.txt             │        after-buffering.txt         │
[...]

PR #71, opened 2023-05-24
Enable breaking lines within "words" by breaking on grapheme clusters

whereswaldon:
This commit implements wrapping on grapheme cluster boundaries (UAX#29) in addition to our previous word boundary (UAX#14) wrapping. The choice of whether to wrap on grapheme clusters is left up to a new LineBreakPolicy type with three options:

- WhenNecessary: This option will break using grapheme clusters when no word can fit on the line or when it is truncating a word (preserving as many grapheme clusters of the word as possible). This is the default policy.
- Always: This option will break on grapheme clusters whenever possible, fitting as many grapheme clusters as possible on every line.
- Never: This option disables the use of grapheme cluster breaking entirely, and preserves the prior behavior.

As WhenNecessary is the zero value of LineBreakPolicy, consumers of our API will experience a behavior change when updating to this version. Text that previously overflowed its bounds because words were too long will now fit within them. I think that this is going to be an improvement for 95% or more of use cases, so the change is warranted.

~This PR is a draft while its prerequisite, #68, is still under review.~

### DIFF
--- a/shaping/fuzz_test.go
+++ b/shaping/fuzz_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/go-text/typesetting/di"
 	"github.com/go-text/typesetting/language"
+	"github.com/go-text/typesetting/segmenter"
 )
 
 // FuzzE2E shapes and wraps large strings looking for unshapable text or failures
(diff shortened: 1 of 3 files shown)


andydotxyz:
> Text that previously overflowed its bounds because words were too long will now fit within them. I think that this is going to be an improvement for 95% or more of use cases, so the change is warranted.

I guess it depends on whether this algorithm is also used to determine space required for text passed.
For example Fyne uses wrapping to know what a minimum would be for width before deciding what space is allocated. If the default was wrap-word then changing the default to be sub-word may not be the expected result.

p.s. what does "UAX#" refer to?

whereswaldon:
> > Text that previously overflowed its bounds because words were too long will now fit within them. I think that this is going to be an improvement for 95% or more of use cases, so the change is warranted.
> 
> I guess it depends on whether this algorithm is also used to determine space required for text passed. For example Fyne uses wrapping to know what a minimum would be for width before deciding what space is allocated. If the default was wrap-word then changing the default to be sub-word may not be the expected result.

Sure, but that isn't the primary or default use case of the algorithm. Fyne is free to ensure that in the APIs it exposes, it sets the policy to `Never` instead of `WhenNecessary`. I still maintain that the API presented by `go-text` is most ergonomic with this default.

> p.s. what does "UAX#" refer to?

These are standards defining how to delimit segments of text put out by unicode:

- [UAX#14](https://unicode.org/reports/tr14/)
- [UAX#29](https://unicode.org/reports/tr29/)

These two documents define the rules for what constitutes a valid line break.

andydotxyz:
> Fyne is free to ensure that in the APIs it exposes, it sets the policy to Never instead of WhenNecessary.

For sure, I wasn't meaning to state that it should be otherwise on Fyne's behalf, just indicating another usage in case it helped.

> I still maintain that the API presented by go-text is most ergonomic with this default.

Fair enough, I don't feel strongly. I think I had misremembered that the default wrapping was set as word - but on review that was more implicit, so the change isn't so large as I had thought.

> These are standards defining how to delimit segments of text put out by unicode:

Thanks

whereswaldon:
I hadn't realized conflicts snuck in on this branch. I should be able to tackle those tomorrow. I suspect they're trivial, so this should be reviewable if you have time.

andydotxyz:
GitHub is showing a few 100 lines of conflict - but it may be just the way they render overlapping code I'm not sure.

benoitkugler (changes requested):
Overall this looks good to me, nice work !
Please have a look at #72 for minor suggestions..

whereswaldon:
What an excellent review @benoitkugler! I really appreciate it.

andydotxyz (approved):

benoitkugler (approved):

PR #72, opened 2023-05-31
Within word review

benoitkugler:
These are my suggestions regarding #71.
I've added some tests and simplify some logic, but there is no fundamental change.
I realize some of the changes could have been proposed earlier when reviewing #68, sorry for delay..


whereswaldon (approved):
A PR with a bunch of improvements is the best kind of code review. Thank you very much! The simplifications and extra tests are most welcome! In particular, I didn't have the distance from the code to see the opportunity to simplify fillUntil, but you're so right that it was overly complex.

PR #77, opened 2023-06-16
Update Harfbuzz version

benoitkugler:
This PR update Harfbuzz to match a recent version of the C library. There is also some minor API cleanup, and minor changes in the unicodedata package.

There should not be any API change for end users.


whereswaldon (approved):
I can only say thank you. :D Keeping up with upstream is important, and this was clearly a lot of work.

andydotxyz (approved):
OMG, good work

PR #79, opened 2023-06-16
Fix cff charstring parsing

benoitkugler:
Properly handle 16.16 fixed floats that may occur in 'cff' tables. 
Related to https://github.com/go-text/render/pull/8


whereswaldon (approved):
Thanks! I assume there's not much to be gained from trying to preserve the fractional component of the fixed?

andydotxyz (approved):
Thanks so much for getting this working

benoitkugler:
> Thanks! I assume there's not much to be gained from trying to preserve the fractional component of the fixed?

@whereswaldon Actually its more that all the interpreter code should be updated to support float instead of integers. I've followed Nigel Tao sfnt implementation decision.

PR #81, opened 2023-06-23
Reduce allocations in metadata

benoitkugler:
Providing a buffer for metadata extractions reduces allocation during system fonts scan.

The API change is required to make sure we don't keep a reference on the shared buffer (whose content may have been overwritten before resolving the family and aspect).

### DIFF
--- a/opentype/api/font/font.go
+++ b/opentype/api/font/font.go
@@ -231,7 +231,8 @@ var bhedTag = loader.MustNewTag("bhed")
 // If a 'bhed' Apple table is present, it replaces the 'head' one.
 //
 // 'buffer' may be provided to reduce allocations; the return Head is guaranteed
-// not to retain any reference on 'buffer'
+// not to retain any reference on 'buffer'.
+// If 'buffer' is nil or has not enough capacity, a new slice is allocated (and returned).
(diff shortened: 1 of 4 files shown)


opentype/api/font/font.go
```diff
@@ -229,22 +229,22 @@ var bhedTag = loader.MustNewTag("bhed")
...
 // If a 'bhed' Apple table is present, it replaces the 'head' one.
-func LoadHeadTable(ld *loader.Loader) (tables.Head, error) {
-	var (
-		s   []byte
-		err error
-	)
+//
+// 'buffer' may be provided to reduce allocations; the return Head is guaranteed
+// not to retain any reference on 'buffer'
+func LoadHeadTable(ld *loader.Loader, buffer []byte) (tables.Head, []byte, error) {
```
andydotxyz, line 235: Should we be explicit for this public API that `nil` will cause a full allocation, but also that if the buffer is too small a new one will be allocated?

benoitkugler: Fair enough, thanks. I've added some documentation.

whereswaldon (approved):
Thanks!

andydotxyz (approved):

PR #85, opened 2023-07-05
Fontscan fixes

whereswaldon:
This PR fixes a couple of small problems I've encountered with the fontscan API:

- If given an empty slice for queried font families, it used fail to find anything
- If we reached the "arbitrary face" return logic, we iterated over a map and returned a random face each iteration


benoitkugler (approved):
Looks good to me, thanks !

andydotxyz (approved):
Nice fixes, sorry for the delay

PR #88, opened 2023-07-07
Script search script set

benoitkugler:
@whereswaldon This is a prototype of the idea I had to build the script set using ranges (for #87 )

I think this approach is exact and runs a bit faster.

However, there is one gotcha : runes not belonging to any script (that is runes with the Unknown script) are not detected.
I think that it is not an issue in practice, but I would like to hear your thoughts.

This is really a draft that I will cleanup if you think this approach make sense.


benoitkugler:
It occurs to me that a rather simple solution for the Unknown script issue would be to add the missing ranges in the ScriptRanges list. It is a bit of a memory waste, but it may be worth trying.

whereswaldon:
I think this approach is much better, yes. :D Also, I think the unknown case is worth handling. How much more memory would that really take?

benoitkugler:
> Also, I think the unknown case is worth handling. How much more memory would that really take?

Good question, I'll do it to see how many item need to be added.

whereswaldon:
> > Also, I think the unknown case is worth handling. How much more memory would that really take?
> 
> Good question, I'll do it to see how many items need to be added.

As a possible alternative, perhaps there's a way to write the loop to express "if the rune range covers a gap between defined scripts, add unknown to the set"? Seems like a space-time tradeoff. Either we add memory to put unknown in the list or we complicate the loop to handle it.

whereswaldon:
That was surprisingly straightforward! I think that solution looks good.

benoitkugler:
Hum, I'm not sure it handles all the cases properly. I'm on the right track though, I'll try to get it right tomorrow!

whereswaldon:
Okay, thank you for the effort and the update!

benoitkugler:
That should be correct now, sleeping helps ! 
If you are happy with this change, we can merge it and conclude review in #87.

whereswaldon (approved):
This looks great; thank you very much. I'll go ahead and merge into script search

PR #89, opened 2023-07-10
[shaping] properly count clusters for vertical text

benoitkugler:
The `Glyph.RunesCount` property for vertical text was always 0, since `countClusters` was not handling vertical directions.

Besides, the `Progression` type makes the fix trivial, thanks for this idea !


whereswaldon (approved):
I'm glad that separating axis and progression is actually making our lives easier. :D I hoped that it would.

andydotxyz (approved):
👍

PR #91, opened 2023-07-10
Support font features

benoitkugler:
This PR add a field on `Input` to support font features.

A test font file with the 'frac' feature is also added.


shaping/shaper.go
```diff
@@ -94,8 +94,18 @@ func (t *HarfbuzzShaper) Shape(input Input) Output {
 	font.XScale = int32(input.Size.Ceil()) << scaleShift
 	font.YScale = font.XScale
 
+	features := make([]harfbuzz.Feature, len(input.FontFeatures))
```
whereswaldon, line 97: Can we make this a reusable buffer on the shaper implementation rather than a newly-allocated buffer per-shaping?

shaping/input.go
```diff
@@ -37,6 +43,12 @@ type Input struct {
 	Language language.Language
 }
 
+// FontFeature sets one font feature.
+type FontFeature struct {
+	Tag   loader.Tag
+	Value uint32
+}
```
whereswaldon, line 50: Could you add a little more documentation about this type? A link to a list of standard font feature tags and common values would be sufficient. As it stands, I don't think many users will understand what this is.

shaping/shaper.go
benoitkugler: Fair enough !

whereswaldon (approved):
Thanks! I've been hoping to surface this in Gio's text API soon.

andydotxyz (approved):
Cool

PR #92, opened 2023-07-12
fontscan: initialize candidates after cache check

whereswaldon:
This commit moves the computation of the font candidates for the current query to after we check for a cache hit. This saves an enormous quantity of redundant work in real applications, as similar faces are frequently requested repeatedly, and the information necessary to resolve them often lives entirely in the cache.

### DIFF
--- a/fontscan/fontmap.go
+++ b/fontscan/fontmap.go
@@ -49,6 +49,8 @@ type FontMap struct {
 	scriptMap map[language.Script][]int
 	lru       runeLRU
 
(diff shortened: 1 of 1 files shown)


benoitkugler:
Very clever. I initially thought the code was not correct when they query has changed, but since the cache actually is query aware, it is not an issue! Nicely done!

benoitkugler (approved):

andydotxyz (approved):

PR #93, opened 2023-07-12
opentype/api/font: use concrete error instead of fmt.Errorf

whereswaldon:
This call to fmt.Errorf is a significant hotspot on Gio's shaping benchmarks. Using a concrete error type makes instantiating this error much cheaper and defers any acutal string formatting until when/if the error is printed.

### DIFF
--- a/opentype/api/font/renderer.go
+++ b/opentype/api/font/renderer.go
@@ -271,10 +271,16 @@ func buildSegments(points []contourPoint) []api.Segment {
 	return out
 }
 
+type errGlyphOutOfRange int
+
+func (e errGlyphOutOfRange) Error() string {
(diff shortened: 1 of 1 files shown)


benoitkugler:
What do you think about returning a "constant" error (droping the reference to the glyph) ?
The function is only used once, and the error value is actually unused... So a constant error is sufficient. And it should solve the performance issue as well.

whereswaldon:
If you prefer that solution, feel free to push it instead. :D I was trying to preserve any debuggability offered by the current error message, but if you don't think that value is useful, a constant error is fine!

benoitkugler (approved):
Let's go with this nice solution, we can always simplify the code further, later.

PR #94, opened 2023-07-12
fontscan: use interface for logger

whereswaldon:
This allows client applications to substitute different logging libraries than simply the stdlib by implementing an adapter that conforms to this interface.


benoitkugler (approved):
Looks good to me !

andydotxyz (approved):
:+1:

PR #97, opened 2023-07-20
Regression fixes

whereswaldon:
This sequence of commits fixes a class of bugs arising from a desynchronization of two elements of line wrapper state. I've included the regression test that first identified the problem, a fix (unifying the manipulation of these related data elements), and a small cleanup commit to make some test case data easier to read.


benoitkugler (approved):
Nicely done, with an increased readability!

andydotxyz (commented):
Good catch!

andydotxyz (approved):
Good catch!

PR #98, opened 2023-08-02
shaping: fix empty string wrapping with width zero

whereswaldon:
This commit ensures that we properly handle wrapping the empty string even when we have zero width to work with. It's common for application code to shape+wrap an empty string to get line metrics, and sometimes they may do that with zero width because the empty string needs no width. However, a faulty comparison in our logic for handling the empty string would fail because 0 is not less than 0. I've added tests triggering the problem and fixed the comparison logic.


andydotxyz (approved):
Nice catch

PR #101, opened 2023-10-12
Fix variations parsing

benoitkugler:
I've discovered a bug when trying to load a corrupted (subsetted) font file. The issue is that we can't really trust the count of point numbers found in the 'glyf' table. 
This commit adds a check to avoid out of bounds crashes.


whereswaldon (approved):

whereswaldon:
I wouldn't complain about adding a regression test for this, but I won't require it.

andydotxyz (approved):
Likewise

benoitkugler:
Test added !

andydotxyz (approved):
Thanks

PR #102, opened 2023-10-13
Opentype font writer

benoitkugler:
This PR adds support to write an Opentype font file from a list of tables. 

This is a simple implementation, since neither font collections nor compressed format (WOFF) are supported.

This paves the way for features such as subsetter or variable font instantiater.

Conceptually, the `Write` function performs the reverse operation of `NewLoader`, so that I think it should belong to this package.

Note that the GUI toolkits are probably NOT interested by such features, but since the Go linker is able to remove dead code, it should come with no cost for end users anyway.


opentype/loader/writer.go
```diff
@@ -0,0 +1,85 @@
...
+// Table is one opentype binary table and its tag.
+type Table struct {
+	Content []byte
+	Tag     Tag
+}
+
+// Write creates a single font file from the given [tables] slice,
+// which must be sorted by Tag
+func Write(tables []Table) []byte {
+	buffer := writeTTF(tables)
```
andydotxyz, line 17: This function has a single purpose - to call `writeTTF`, which is strange - if it is generic we don't need the extra function, but if it is because this writer is for TTF perhaps the generic name isn't right?

andydotxyz (commented):
Cool addition - not quite sure about the naming though - will this support other font types like collections? if so the generic `Write` name may need clarity.

opentype/loader/writer.go
benoitkugler: Fair point. In the future, we may want to support other formats, like web font (WOFF). In such case, the Write function would become `Write(tables []Table, woff bool) []byte`, and internally call `writeTTF` or `writeWOFF`. 
But, for now, it seems indeed clearer to drop the internal function and rename `Write` to, say, `WriteTTF`.

andydotxyz (approved):

whereswaldon (approved):
Sorry for the delay. Feature looks good! Thank you.

PR #103, opened 2023-10-16
Better space handling ?

benoitkugler:
Following the discussion on #90 , this is a try at better handling trailing spaces.

This is not complete (tests are not event passing), but I would be interested in early feedback.

One of the difficulty is about handling truncation properly :
I think the behavior should be, for a width of 5, to truncate
`quick_fox`
to 
`quick_...` (but with a zero advance space)
and not to 
`quick...`

### DIFF
--- a/shaping/output.go
+++ b/shaping/output.go
@@ -128,6 +128,29 @@ func (o *Output) RecomputeAdvance() {
 	o.Advance = advance
 }
 
+// advanceSpaceAware adjust the value in [Advance]
+// if a white space character end the run.
+// TODO: should be take into account multiple spaces ?
+func (o *Output) advanceSpaceAware() fixed.Int26_6 {
+	L := len(o.Glyphs)
+	if L == 0 {
+		return o.Advance
+	}
+
+	// adjust the last to account for spaces
+	if o.Direction.IsVertical() {
+		if g := o.Glyphs[L-1]; g.Height == 0 {
+			return o.Advance - g.YAdvance
+		}
+	} else { // horizontal
+		if g := o.Glyphs[L-1]; g.Width == 0 {
+			return o.Advance - g.XAdvance
+		}
+	}
+
+	return o.Advance
+}
+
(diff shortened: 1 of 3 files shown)


whereswaldon (commented):
Thank you for tackling this! I think this approach is sound conceptually, and I agree that it's best to keep trailing zero-width space at the end of lines than to push it to the start of new lines.

We should eventually handle the case of multiple consecutive whitespaces at the end of a run, but I think it's reasonable to start with just one.

I was surprised that we're detecting whitespace purely based on whether the Width/Height fields of the glyph metadata are zero. In practice, I suppose that should be effective. I just expected something fancier.

benoitkugler:
@whereswaldon I've adjusted some tests in `TestLineWrapperBreakSpecific`. I think the new output values are fine, but I also think some of the tests may not be as informative as previously, with respect to the `BreakPolicy` option. When you have some time, would it be possible for you to carefully review these changes ?

whereswaldon:
> @whereswaldon I've adjusted some tests in `TestLineWrapperBreakSpecific`. I think the new output values are fine, but I also think some of the tests may not be as informative as previously, with respect to the `BreakPolicy` option. When you have some time, would it be possible for you to carefully review these changes ?

@benoitkugler Done! I've fixed up two tests to ensure that they still check the right things. Looks good to me now.

benoitkugler:
> @benoitkugler Done! I've fixed up two tests to ensure that they still check the right things. Looks good to me now.

Many thanks !

whereswaldon (approved):
Thanks for tackling this important feature! It's a big deal for centering and right-aligning LTR text.

andydotxyz:
Looks great. Wish I had time to test replacing our white space / align code from Fyne but I have other open PRs that clash at the moment.

andydotxyz (approved):
Looks great. Wish I had time to test replacing our white space / align code from Fyne but I have other open PRs that clash at the moment.

PR #104, opened 2023-11-09
[fontscan] Add support for lang set

### LINKED ISSUE #95: Implement font language coverage and querying
This has been discussed previously (see below the hr)

I've discovered what I think is a compelling use case for including language info in footprints: determining the "primary" font of a piece of text. I'd like to set the default line height for a paragraph to the builtin line height of the primary font in the text. However, it's difficult to define "primary" font because you can't know *a priori* what font will be used. It's chosen based on aspect and the codepoints in the text.

You could consider using heuristics like the most-frequently-occurring font within the text, but you can also trivially create pathological cases that defeat such logic.

The best option I've been able to devise is this:

> the "primary" font is the font that will be selected for the given query term when shaping the user's system language (in a UI context)

This should result in a stable choice of primary font that will work well with the rest of an application's UI.

Implementing this is tricky as there isn't a good mapping between system `language.Language` and `language.Script` (and as far as I understand, there cannot be). For that reason, I think the only way to query "which font will be used for this query when displaying the system language" is to expose a FontMap `ResolveFace`-like operation that acts on languages instead of runes. This, in turn, requires our footprints to carry supported languages so that they can be efficiently queried.

@benoitkugler Does this make sense and seem like a good approach? I'm happy to work on this feature, or split the work between us if you have a concrete idea of how to implement parts of this.

---

> By the way, I think there is one change we may safely forecast : adding the set of languages supported by a font. It may be deduced from the `runeSet`, but, similarly to the `scriptSet`, is quite expensive to compute, so I guess we would rather store it on the footprint.
> 
> I should be able to post a PR soon if you think we should add it before the index format is widely used.

_Originally posted by @whereswaldon in https://github.com/go-text/typesetting/issues/87#issuecomment-1629478341_

benoitkugler:
This PR adds support for computing the languages supported by a font, using the orthographic database from fontconfig.

Note that the cache version had to be dumped since the lang set is added in the footprint fields.

There is (almost) no exported API yet : to close #95, we probably want to add a `ResolveForLanguage(language LanguageID) Face` method on the `FontMap`. Should we add it in this PR ?


whereswaldon:
> This PR adds support for computing the languages supported by a font, using the orthographic database from fontconfig.
> 
> Note that the cache version had to be dumped since the lang set is added in the footprint fields.
> 
> There is (almost) no exported API yet : to close #95, we probably want to add a `ResolveForLanguage(language LanguageID) Face` method on the `FontMap`. Should we add it in this PR ?

I haven't had time to review this in depth yet, but I do think we should surface this feature in the same PR. Otherwise, we're dropping the cache version for users without offering them a new capability in exchange. Note that I haven't actually looked at the public API changes, so I'm speaking in ignorance. That's just my gut reaction to your question. I'll try to review properly soon.

benoitkugler:
I've added `ResolveFaceForLang` as the external API, leveraging the langset built on the footprints.
This method may return nil if the language is not supported by the fontmap.

Typical usage would be : 
```go
langID, ok := NewLangID("en-us")
if !ok {
    // the language is not even supported by the library: abort
    return 
}
face := fm.ResolveFaceForLang(langID)
if face == nil {
    // the fontmap does not support the given language: abort 
    return 
}
// all good : use face

File: fontscan/langset.go [hunk 1]
```diff
@@ -0,0 +1,22 @@
+package fontscan
+
+// LanguageID is a compact representation of a language
+// this package has orthographic knowledge of.
+type LanguageID uint16
```
whereswaldon: If I'm reading the code right, we're only using 9 bits of the 16. That's probably worth documenting somewhere. Maybe on the `langset` itself, since it's talking about the "high bits" being bits 8-6.

fontscan/langset.go
```diff
@@ -1,14 +1,85 @@
...
+)
+
 // LanguageID is a compact representation of a language
 // this package has orthographic knowledge of.
 type LanguageID uint16
 
+// NewLanguageID returns the compact index of the given language,
+// or false if it is not supported by this package.
+//
+// Derived language no exactly supported are mapped to their primary part : for instance,
```
whereswaldon, line 18: `s/language no/languages not/`

fontscan/rune_coverage.go
```diff
@@ -278,7 +278,7 @@ func (a runeSet) Len() int {
 
 const runePageSize = 2 + 8*4 // uint16 + 8 * uint32
 
-// serialize serialize the Coverage in binary format
+// serialize serialize the rune coverage in binary format
```
whereswaldon, line 281: `s/serialize the/serializes the/`

fontscan/fontmap.go
```diff
@@ -413,6 +412,25 @@ func (fm *FontMap) resolveForRune(candidates []int, r rune) font.Face {
 	return nil
 }
 
+// returns nil if not candidates supports the language `lang`
```
whereswaldon, line 415: `s/if not candidates supports/if no candidates support/`

fontscan/fontmap.go
```diff
@@ -413,6 +412,25 @@ func (fm *FontMap) resolveForRune(candidates []int, r rune) font.Face {
...
+func (fm *FontMap) resolveForLang(candidates []int, lang LanguageID) font.Face {
+	for _, footprintIndex := range candidates {
+		// check the coverage
+		if fp := fm.database[footprintIndex]; fp.langs.contains(lang) {
+			// try to use the font
+			face, err := fm.loadFont(fp)
+			if err != nil { // very unlikely; try an other family
```
whereswaldon, line 422: `s/an other/another/`

whereswaldon (approved):
Great work! It all makes sense to me. I've called out some typos and an opportunity to clarify the docs on the binary representation, but none of that is a blocker.

andydotxyz (approved):
Likewise. I'm not sure I understand it but the docs are clear and it has tests :)

whereswaldon:
I'll fix up the typos myself and then merge.

whereswaldon:
Actually, no. I'll fix up the typos, but wait to hear @benoitkugler's thoughts on changing the binary representation docs before merge.

benoitkugler:
Thank you for the review and the typos fix.

I've updated `langset` documentation following your advice. Feel free to alter it, or to merge if you are happy with that.

whereswaldon (approved):
Thanks for the doc update!

PR #105, opened 2023-11-17
[fontscan] Better family substitutions

benoitkugler:
This PR alters the way we perform family substitutions when looking for a fallback font, by applying them to the whole family list of the user query, so that substitution priorities are better respected.

Let me illustrate the change on the following example, where an user queries `{CustomWeirdFont, Helvetica}`, on a system with neither `CustomWeirdFont` nor `Helvetica`, and calls `ResolveFace('a')`

### Before the change

No exact match occurs (step 1), so we apply substitutions (step 2). `CustomWeirdFont` is unknown from the library rules, so it is expanded by default to `sans-serif` for which the system has (many) fonts (with support for the rune 'a'). So `ResolveFace` returns the 'preferred' `sans-serif` face. It has not take into account the `Helvetica` substitutions.

### After the change 

We still look for an exact match, in the user query order. Like before, no one occurs. We now apply substitutions on the whole  `{CustomWeirdFont, Helvetica}` list. Since the library has rules for `Helvetica`, the expanded families will look like `{CustomWeirdFont, Helvetica, NimbusSans, ..., other sans-serif families}`, with the main difference being that `NimbusSans`, which is the preferred substitution for `Helvetica`, has now higher priority than the default sans-serif substitutions coming from `CustomWeirdFont`.

This behavior is closer to what fontconfig does on Linux systems, and seems overall a better way of looking for fallbacks.

### DIFF
--- a/fontscan/fontmap.go
+++ b/fontscan/fontmap.go
@@ -325,30 +325,27 @@ func (fm *FontMap) SetQuery(query Query) {
 }
 
 // candidates is a cache storing the indices into FontMap.database of footprints matching a Query
+// families
 type candidates struct {
-	// the two fallback slices have the same length: the number of family in the query
-	withFallback    [][]int // for each queried family
-	withoutFallback []int   // for each queried family, only one footprint is selected
+	// footprints with exact match :
+	// for each queried family, at most one footprint is selected
+	withoutFallback []int
+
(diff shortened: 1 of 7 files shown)


andydotxyz (approved):
Description sounds great and the code looks solid too, nice one :)

whereswaldon (approved):
Thank you; this is a definite improvement.

PR #106, opened 2023-11-17
[harfbuzz] reuse otApplyContext instead of allocating a new one

benoitkugler:
This makes `Buffer` more memory efficient when reusing the same font, since the `otMap` is stored on the buffer `planCache` .


andydotxyz (approved):
Cunning :)

whereswaldon (approved):
:tada:

PR #107, opened 2023-11-18
[fontscan] Expose a way to query a system font by family

benoitkugler:
CSS [@font-face rule](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face) allows a document author to query a system (or local) font by _family name_. Although similar to the functionnality `SetQuery` and `ResolveFace` are providing, this is a different task (no rune involved, no aspect involved, only the system fonts are used).  Also, no fallback is desirable.

This PR then provides a `FindSystemFont(family string)` (its implementation is rather simple with our internal index).

### DIFF
--- a/fontscan/fontmap.go
+++ b/fontscan/fontmap.go
@@ -314,6 +314,26 @@ func (fm *FontMap) FontMetadata(ft font.Font) (family string, aspect meta.Aspect
 	return item.Family, item.Aspect
 }
 
+// FindSystemFont looks for a system font with the given [family],
+// returning the first match, or false is no one is found.
+//
+// User added fonts are ignored, and the [FontMap] must have been
+// initialized with [UseSystemFonts] or this method will always return false.
+//
+// Family names are compared through [meta.Normalize].
+func (fm *FontMap) FindSystemFont(family string) (Location, bool) {
+	family = meta.NormalizeFamily(family)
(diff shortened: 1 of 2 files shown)


fontscan/fontmap.go
```diff
@@ -314,6 +314,26 @@ func (fm *FontMap) FontMetadata(ft font.Font) (family string, aspect meta.Aspect
...
 }
 
+// FindSystemFont looks for a system font with the given [family],
+// returning the first match, or false is no one is found.
+//
+// User added fonts are ignored, and the [FontMap] must have been
+// initialized with [UseSystemFonts] or this method will always return false.
+//
+// Family names are compared through [meta.Normalize].
+func (fm *FontMap) FindSystemFont(family string) (Location, bool) {
```
andydotxyz, line 324: Just a question - is the not found case better as "<undefined> false" or "nil" to avoid the second return parameter?

benoitkugler: No strong here, but, since Location is a struct, that would force the user to check Location.File against the empty string, which I find a bit akward. Also, the current signature matches the general map lookup syntax in Go.

whereswaldon: I like matching the semantics of map access for this. An explicit value signalling whether anything was found.

whereswaldon (approved):

PR #108, opened 2023-11-21
[opentype/metadata] Fix aspect resolution

benoitkugler:
Previous to this change, we were not using the 'OS/2' table when computing the `Aspect` of a font. 

It is actually an outstanding issue, but it went undetected since the fallback `inferFromStyle` was doing a pretty good job. 

However, there is (at least) one font (DejaVu Sans) which has different values for its weight (that is, the OS/2 and style string values do not match). In this case, it is better to give priority to the 'OS/2' table (at least, that is what fontconfig does).

Also, I've added the font file in the merged https://github.com/go-text/typesetting-utils/pull/14 PR.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,5 +4,5 @@ go 1.17
 
 require (
 	github.com/go-text/typesetting-utils v0.0.0-20231121125213-61dadda55b86
-	golang.org/x/image v0.14.0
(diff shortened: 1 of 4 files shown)


whereswaldon (approved):
Nice catch!

andydotxyz (approved):

Issue #109, opened 2023-11-22
About input segmentation


benoitkugler:
I need to perform input segmentation, in a similar way gioui is [doing](https://git.sr.ht/~eliasnaur/gio/tree/main/item/text/gotext.go#L423). 
 
This issue is to discuss how to include such a feature in go-text, if you are OK with it.

What would you think of adding a new `shaping.Segmenter` type to perform the three segmentations : bidi, by face coverage, by script ? 
It could internally use a single buffer of `Input`s, so that it is memory efficient but provides a clean API.

The current `SplitByFace` function could be deprecated/replaced/re-implemented with this new type.

<details><summary>Side note regarding the implementation of the script segmenter </summary>
<p>

We could perhaps enhance a lit bit the one in gioui, to better handle Common script. Following the advice at https://unicode.org/reports/tr24/#Common, what Pango does is keeping track of parenthesis and give closing delimiters the script of their open counterpart.

</p>
</details>

PR #110, opened 2023-11-24
[shaping] Add an input segmenter

benoitkugler:
This is prototype for #109 

I propose the following API 
 
```go
// Split segments the given [text] according to :
//   - text direction
//   - script
//   - face, as defined by [faces]
//
// As a consequence, it sets the following fields of the returned runs :
//   - Text, RunStart, RunEnd
//   - Direction
//   - Script
//   - Face
//
// [defaultDirection] is used during bidi ordering, and should refer to the general
// context [text] is used in (typically the user system preference for GUI apps.)
//
// The returned sliced is owned by the [Segmenter] and is only valid until
// the next call to [Split].
func (seg *Segmenter) Split(text []rune, faces Fontmap, defaultDirection di.Direction) []Input
```

I'm happy to read your thoughts !


benoitkugler:
> What would you think about changing the API to:

Done in [2f2ea8d](https://github.com/go-text/typesetting/pull/110/commits/2f2ea8d97e5bf4c9f837eb5ddf61c98e63511e45). I've remove the `defaultDirection` parameter, since we can now use the `Direction` field of the `Input`.

whereswaldon (approved):
Looks excellent! Thank you very much for the work on this.

andydotxyz (approved):
Thanks. I don't have time to test the functionality in Fyne yet, but it looks good and tests are in anyway :)

PR #114, opened 2023-11-26
opentype, fontscan: remove go1.16 build constraints

hajimehoshi:
Now this project uses Go 1.17 and newer, an old code for Go 1.15 and older should not be needed.


hajimehoshi:
I assumed this project assumpts Go 1.17 from go.mod, but perhaps do we need to test this with Go 1.14?

benoitkugler:
Thank you for this contribution.

> I assumed this project uses Go 1.17 from go.mod, but perhaps do we need to test this with Go 1.14?

It was indeed the case, but now that Fyne has bumped its minimum version, I think it should be OK. 
Cc @andydotxyz for confirmation

andydotxyz (approved):
Looks like a good catch thanks

benoitkugler (approved):
Nice clean up !

PR #115, opened 2023-11-27
[shaping/wrapper] return the index of the next line in WrapLine

benoitkugler:
This PR also returns the internal value used by the line wrapper, which is somewhat annoying to compute, but very useful when processing a text line by line.


andydotxyz (approved):
Looks good, sorry for the delay

PR #117, opened 2023-11-28
[opentype/gpos] use plain struct instead of pointer

benoitkugler:
During some benchmarks, I found `tables.PairSet.FindGlyph` was a really high source of heap allocations.
Using a plain struct (probably) allows the compiler/runtime to allocate it on the stack, resulting in a sizable reduction of heap allocations.

### DIFF
--- a/harfbuzz/ot_layout_gpos.go
+++ b/harfbuzz/ot_layout_gpos.go
@@ -285,8 +285,8 @@ func (c *otApplyContext) applyGPOSPair1(inner tables.PairPosData1, index int) bo
 	skippyIter := &c.iterInput
 	pos := skippyIter.idx
 	set := inner.PairSets[index]
-	record := set.FindGlyph(gID(buffer.Info[skippyIter.idx].Glyph))
-	if record == nil {
+	record, ok := set.FindGlyph(gID(buffer.Info[skippyIter.idx].Glyph))
+	if !ok {
 		buffer.unsafeToConcat(buffer.idx, pos+1)
 		return false
 	}
--- a/opentype/tables/ot_properties.go
+++ b/opentype/tables/ot_properties.go
@@ -215,13 +215,13 @@ func (lk ExtensionPos) Cov() Coverage         { return nil } // not used anyway
 
 // FindGlyph performs a binary search in the list, returning the record for `secondGlyph`,
 // or `nil` if not found.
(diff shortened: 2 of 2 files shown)


whereswaldon:
@benoitkugler Do you have before/after benchmark info you could share?

benoitkugler:
> @benoitkugler Do you have before/after benchmark info you could share?

Here you are : the code simply runs Shaper.Shape, with a font using a GPOS table (Arimo).

<details><summary>Code</summary>
<p>

	face := loadOpentypeFont(b, "/usr/share/fonts/truetype/croscore/Arimo-Regular.ttf")
	textInput := []rune("Une jolie phrase en français !")
	input := Input{
		Text:      textInput,
		RunStart:  0,
		RunEnd:    len(textInput),
		Direction: di.DirectionLTR,
		Size:      16 * 72,
		Face:      face,
		Script:    language.Latin,
		Language:  language.NewLanguage("fr"),
	}
	shaper := HarfbuzzShaper{}

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		_ = shaper.Shape(input)
	}

</p>
</details> 

### Benchmark stats
   
```
       │   old.txt   │              new.txt               │
       │   sec/op    │   sec/op     vs base               │
GPOS-4   40.91µ ± 2%   37.47µ ± 5%  -8.41% (p=0.000 n=10)

       │   old.txt    │               new.txt                │
       │     B/op     │     B/op      vs base                │
GPOS-4   15.94Ki ± 0%   11.88Ki ± 0%  -25.48% (p=0.000 n=10)

       │  old.txt   │              new.txt               │
       │ allocs/op  │ allocs/op   vs base                │
GPOS-4   253.0 ± 0%   227.0 ± 0%  -10.28% (p=0.000 n=10)
``` 

### Memory profile - Before: 
old

### Memory profile - After : 
new

andydotxyz:
Has the "After" trimmed the RunN contribution or did that go to 0?
I can't do the maths accurately on the fly but it looks like memory is just being moved to other places rather than reduced overall.

benoitkugler:
> Has the "After" trimmed the RunN contribution or did that go to 0? I can't do the maths accurately on the fly but it looks like memory is just being moved to other places rather than reduced overall.

My screenshot was incomplete, see the new one : 

Capture d’écran du 2023-11-29 12-15-00

whereswaldon (approved):

andydotxyz (approved):
Thanks for the clarification :)

Issue #112, opened 2023-11-24
[fontscan] Provide precise Location in FontMap.AddFace


benoitkugler:
I use `FontMap.AddFace` to have control about the description registered with a font, but I would also control the file identifier.
That is, I would like to also pass a `Location`. For now, the package generate a synthetic file name with 
`out.Location.File = fmt.Sprintf("%v", md)` (`md` being the font description provided by the caller).

Would you be OK to change the signature to `AddFace(face font.Face, id Location, md meta.Description)` ?

The previous behavior could be easily obtained with 
`AddFace(face font.Face, Location{File: fmt.Sprintf("%v", md)}, md meta.Description)`

Pinging @whereswaldon as code author

whereswaldon:
I have no objection; do it!

PR #119, opened 2023-12-02
[fontscan] Specify Location in AddFace

benoitkugler:
Let the user choose the `Location` when adding a Face (for #112).


whereswaldon (approved):
Nice and straightforward :D

whereswaldon:
@benoitkugler For future reference, any *breaking* change to the public API does need multiple reviewers, even if it is trivial. It will make downstream projects fail to compile against our `main` branch. If we'd implemented this by adding an alternative method, I think I'd feel differently.

Issue #118, opened 2023-11-30
`NotoSansJP-VF.otf` (OpenType) didn't work, while `NotoSansJP-VF.ttf` (TrueType) worked


hajimehoshi:
https://github.com/notofonts/noto-cjk/releases/tag/Sans2.004

I tested `NotoSansJP-VF.ttf` in "All Variable TTF/OTC" and `NotoSansJP-VF.otf` in "All Variable OTF/OTC", and apparently the OpenType version didn't work. go-text couldn't render any glyphs. I found this in my game engine using go-text, so it is a little hard to create a minimized test case, but this is just a quick note.

Confirmed version of go-text: v0.0.0-20231120180320-af78120ccb13

Thanks,

benoitkugler:
Thank you for the bug report.

Could elaborate on "couldn't render any glyphs" ? Is that a visual problem or are the runs (`Output`) also corrupted ? 
You could perhaps extract an input string so that we can investigate ?

hajimehoshi:
Nothing was rendered actually. Apparently no glyphs were obtained for any texts. As I said, this happened in my game engine, so I'll write down a reproducible code with this engine. I'll take to make a minimal code without the engine to reproduce the issue later.

```go
package main

import (
	"bytes"
	_ "embed"

	// The version is 6a7688d3ac4b6fdc264758885465f20833848c10
	"github.com/hajimehoshi/ebiten/v2/text/v2"
)

//go:embed NotoSansJP-VF.otf
var f []byte

func main() {
	src, err := text.NewGoTextFaceSource(bytes.NewReader(f))
	if err != nil {
		panic(err)
	}
	glyphs := text.AppendGlyphs(nil, "abcあいう", &text.GoTextFace{
		Source: src,
		Size:   10,
	}, nil)
	println(len(glyphs)) // 0 was printed unexpectedly. This means no glyphs were obtained.
}
```

hajimehoshi:
Here is the minimal reproducible test case:

```go
package main

import (
	"bytes"
	_ "embed"
	"fmt"

	"github.com/go-text/typesetting/di"
	"github.com/go-text/typesetting/opentype/api/font"
	"github.com/go-text/typesetting/opentype/loader"
	"github.com/go-text/typesetting/shaping"
	"golang.org/x/image/math/fixed"
)

//go:embed NotoSansJP-VF.otf
var noto []byte

func main() {
	l, err := loader.NewLoader(bytes.NewReader(noto))
	if err != nil {
[...]

PR #120, opened 2023-12-04
Add support for the CFF2 Opentype table

benoitkugler:
This PR adds support for the 'CFF2' Opentype table (required by #118).

It is larger than expected since I had to fix variation handling : we were previously using float32 to store a coordinate, but the Opentype spec uses a fixed 2.14 storage. The difference was significant enough to have a off-by-one difference with Harfbuzz when computing glyph extents (with variations). It seems simpler and better to just follow the spec (and Harfbuzz). 

Also, I had to tweak the Postscript interpreter implementation to use float64 instead of int32, so that we properly handle the non integer point coordinates arising when using variations.


whereswaldon (approved):
I live in awe of your contributions to the typesetting project @benoitkugler. Thank you! Yet again, excellent work. This looks reasonable to me. I do not understand the details of the opentype spec well enough to do deep review, but what I did grok looks great.

andydotxyz (approved):
Likewise :)

PR #121, opened 2023-12-05
[shaping] Document glyph metrics and fix vertical line thickness

benoitkugler:
When exploring how to implement vertical text layout, I had a hard time remembering the meaning of the various glyph metrics. 

So I've written a small utility function displaying an `Output` glyph bounds onto an `image.Image`. I you run the `ExampleShaper_Shape`, you get two images displaying the glyph extents in green square.

Using this tool, I've found what I think is an incorrect glyph bounds calculation for vertical text. 

So this PR is mainly about documenting the various fields of `Glyph` and make sure we are on the same page (it could also be me not using using the proper conventions...).


shaping/output.go
```diff
@@ -11,14 +11,31 @@ import (
...
+	// the glyph content, typically positive
 	XBearing fixed.Int26_6
+	// YBearing is the distance between the dot (with offset applied) and
+	// the top of the glyph content, typically positive
 	YBearing fixed.Int26_6
+	// XAdvance is the distance between the current dot (without offset applied) and the next dot.
+	// It is typically positive for horizontal text, and always zero for vertical text.
 	XAdvance fixed.Int26_6
+	// XAdvance is the distance between the current dot (without offset applied) and the next dot.
+	// It is typically negative for vertical text, and always zero for horizontal text.
```
whereswaldon, line 32:
```suggestion
	// YAdvance is the distance between the current dot (without offset applied) and the next dot.
	// It is typically negative for vertical text, and always zero for horizontal text.
```

whereswaldon (approved):
Thank you very much for the extra clarification. I've struggled with this myself quite a bit.

andydotxyz (approved):
Thanks so much, I think this finally nails an issue that I was trying to understand too!

Issue #111, opened 2023-11-24
How to render vertical texts


hajimehoshi:
Hi, I'm using this library for [my game engine](https://ebitengine.org). This is awesome!

I have several questions about vertical texts:

* How can I render a vertical Mongolican text? Usually Mongolian font files include glyphs for horizontal texts, so I think I have to rotate glyphs by myself based on [a Unicode table for Vertical_Orientation](https://www.unicode.org/Public/UCD/latest/ucd/VerticalOrientation.txt). Is that correct?

(The left-bottom texts are in Mongolian. This rendering result is without Vertical_Orientation)
![image](https://github.com/go-text/typesetting/assets/16950/44b37a4a-6de5-4315-adb3-7af82572d493)

* Even if I rotate glyphs, the Y advances seems too big and glyphs were rendered in incorrect positions. How can I get correct Y advances?

(This rendering result is with Vertical_Orientation)
![image](https://github.com/go-text/typesetting/assets/16950/b7e91710-bc9b-4143-ae2e-094739722e20)

* When I use the Unicode table, "… (U+2026)"'s Vertical_Orientation value is also R, so this glyph seems to have to be rotated. However, Japanese fonts usually have a vertical glyph for this code point, so I don't want to rotate this glyph. You can see the glyph "…" was incorrectly rendered in the above Japanese text (the right-bottom text). Is there a way to know whether a glyph ID is for vertical texts or not?

Thanks!

hajimehoshi:
Rendering a Mongolian text in an horizontal direction seems fine by the way

![image](https://github.com/go-text/typesetting/assets/16950/411f594f-d960-4673-b6aa-68267a78e623)

benoitkugler:
Hi ! 
Thank you for your interest in this package. 

I think rendering vertical text has not been tested thoroughly yet, but it is definitively an area we would like to improve !
I'll dig deeper in your issue; thanks for the detailed problem statement.

PS : As an aside, I've skimmed your project issue tracker, and found https://github.com/hajimehoshi/ebiten/issues/788. Perhaps you will be interested by the `fontscan` (and its `FontMap` type) that you could use to access system fonts.

hajimehoshi:
Thank you for your quick response!

> Perhaps you will be interested by the fontscan (and its FontMap type) that you could use to access system fonts.

Thanks, I'll take a look!

benoitkugler:
From a first research, I've understood that : 

[...]

PR #124, opened 2023-12-08
Better support for vertical text

benoitkugler:
This PR aims at fixing #111, by supporting "sideways" glyph orientation for vertical text.

It is roughly divided in two parts : 

1. Adding a "sideways" flag in `Direction`, and implementing "sideways" mode in `Shape`. A convenience method is provided : [`GlyphOutline.Sideways`](https://github.com/go-text/typesetting/blob/vertical-scripts/opentype/api/font.go#L110) to be used by renderers to rotate glyph.
2. Resolving the preferred orientation base on Unicode rules : this is done in [`Segmenter.Split`](https://github.com/go-text/typesetting/blob/vertical-scripts/shaping/input.go#L153)

The only change required for users/renderers is thus to check against the [`Ouput.Direction.IsSideways()`](https://github.com/go-text/typesetting/blob/vertical-scripts/di/direction.go#L105) boolean, to properly rotate glyphs when rendering.

Happy to read your thoughts !

(@hajimehoshi Maybe you would like to test this change ? The code in `ExampleShaper_Shape` and `drawVGlyphs` may be useful to do so)


hajimehoshi:
Thanks! I tested Monglian and Japanese texts. Fonts are [Noto Sans Monglian](https://fonts.google.com/noto/specimen/Noto+Sans+Mongolian) and [M+ 1 Regular](https://github.com/coz-m/MPLUS_FONTS/tree/master/fonts/ttf).

`SetSideways(false)`:
image

`SetSideways(true)`:
image

* Mongolian texts are not rendered correctly in both cases.
* With `SetSideways(true)`, horizontal glyphs in Japanese like "ー" and "…" are now used. This is not correct as a vertical Japanese text. I am not sure this is intended or not. Also, alphabets are not rotated in both cases.

This is just a quick report. I'll try to make a minimal test case if needed.

hajimehoshi:
I'll test those tomorrow. Could you tell me what APIs were newly added?

benoitkugler:
> I'll test those tomorrow. Could you tell me what APIs were newly added?

So I've updated the [PR description](https://github.com/go-text/typesetting/pull/124#issue-2032961205) with links towards the new API : `Segmenter.Split`, `Direction.IsSideways`, `GlyphOutline.Sideways`. Hope it is clearer now :)

hajimehoshi:
> // If [yOffset] is zero, the resulting glyph will typically be displayed
> // under the baseline. To position the glyph over the baseline, pass
> // the YAdvance as [yOffset] (a positive value to lift the glyph up).

yOffset is float32 while YAdvance is fixed.Int26_6. Should I convert this appropriately?

hajimehoshi:
I utilized `Segment.Split` and `Glyph.Sideways` without `Direction.SetSideways`, and the result seems much better.

image

The Monglians seems fine! I found that mixing Japanese characters and English characters doesn't work well. In the above figure "あHello, World.あ"'s Y offsets does match. The last あ overlaps the adjacent English sentence. Should I adjust this with `HAscent` for each outputs?

Thanks!

benoitkugler:
> > // If [yOffset] is zero, the resulting glyph will typically be displayed
> > // under the baseline. To position the glyph over the baseline, pass
> > // the YAdvance as [yOffset] (a positive value to lift the glyph up).
> 
> yOffset is float32 while YAdvance is fixed.Int26_6. How should I convert this appropriately?

Since YAdvance has been scaled by Input.Size (output = font_units * Size / face.Upem), and Sideways expect a value in font units, you should pass YAdvance / (Size * face.Upem) to convert back to fonts units. More precisely, you could use
`float32((-YAdvance / Size).Round()) / float32(face.Upem())`
which is a bit inaccurate due to Rounding. A better way would be to use a fixed.Int26_6 to float32 conversion : 
```go
func toFloat32(a fixed.Int26_6) float32 {
    return float32(a) / 64
}
```

hajimehoshi:
> A better way would be to use a fixed.Int26_6 to float32 conversion :

OK, I actually did this:

```go
func fixed26_6ToFloat32(x fixed.Int26_6) float32 { 
    return float32(x>>6) + float32(x&((1<<6)-1))/float32(1<<6)
}
```

hajimehoshi:
> Since YAdvance has been scaled by Input.Size (output = font_units * Size / face.Upem), and Sideways expect a value in font units,

So, the simple conversion like `YAdvance / 64` doesn't sound correct in terms of unit conversion, but is this actually better...?

hajimehoshi:
I think I was misunderstanding. Your intention is like this?

```go
fixed26_6ToFloat32(-YAdvance) / Size / float32(face.Upem())
```

hajimehoshi:
I'm sorry for spamming, but

> Since YAdvance has been scaled by Input.Size (output = font_units * Size / face.Upem), and Sideways expect a value in font units, you should pass YAdvance / (Size * face.Upem) to convert back to fonts units.

YAdvance is a value in the scaled units, while Sideways expects a value in the font units. So

```
YAdvance / Size * face.Upem
```

seems correct, but am I missing something?

benoitkugler:
> YAdvance is a value in the scaled units, while Sideways expects a value in the font units. So
> 
> ```
> -YAdvance / Size * face.Upem
> ```
> 
> seems correct, but am I missing something?

Well, to go from font to scaled you multiply by Size / face.Upem; so to revert the process and go from scaled to font you have to divide by (Size / face.Upem)

> YAdvance is a value in the scaled units, while Sideways expects a value in the font units. So
> 
> ```
> -YAdvance / Size * face.Upem
> ```
> 
> seems correct, but am I missing something?

I agree with this formula. Hum... I'll try to render the glyphs from your example and investigate tomorrow.

hajimehoshi:
I'll check this later, but

```
glyphData.Sideways(float32(-g.YAdvance.Round()) / sizeFactor)
```

Isn't the value almost 0?

shaping/shaping_test.go
```diff
@@ -557,6 +552,63 @@ func TestCFF2(t *testing.T) {
...
 }
 
+func TestShapeVerticalScripts(t *testing.T) {
+	b, _ := td.Files.ReadFile("common/NotoSansMongolian-Regular.ttf")
+	monF, _ := font.ParseTTF(bytes.NewReader(b))
+	b, _ = td.Files.ReadFile("common/mplus-1p-regular.ttf")
+	japF, _ := font.ParseTTF(bytes.NewReader(b))
+
+	monT := []rune("ᠬᠦᠮᠦᠨ ᠪᠦᠷ ᠲᠥᠷᠥᠵᠦ ᠮᠡᠨᠳᠡᠯᠡᠬᠦ")
+	japT := []rune("もつ青いそら…")
```
hajimehoshi, line 562: "もつ青いそら…" doesn't make sense as a Japanese text, and "青い空…" is much better if we have.

("あHello, World.あ" also doesn't make sense, but I think it is fine to indend a delebarate sentence mixed in English and Japanese)

benoitkugler: Thank you, I've updated the test.

shaping/render_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+		// draw the dot
+		drawPoint(img, dot, black)
+
+		// draw a sketch of the glyphs
+		glyphData := out.Face.GlyphData(g.GlyphID).(api.GlyphOutline)
+		if out.Direction.IsSideways() {
+			glyphData.Sideways(float32(-g.YAdvance.Round()) / sizeFactor)
+		}
+		drawGlyph(img, dot.Add(image.Pt(g.XOffset.Ceil(), g.YOffset.Ceil())), glyphData, sizeFactor, black)
```
hajimehoshi, line 175: Question: why isn't this `dot.Add(image.Pt(g.XOffset.Ceil(), -g.YOffset.Ceil()))`?

shaping/render_test.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	return dot
+}
+
+// assume vertical direction
+func drawVRun(out Output, img *image.RGBA, dot image.Point) image.Point {
+	sizeFactor := float32(out.Size.Round()) / float32(out.Face.Upem())
+	for _, g := range out.Glyphs {
+		dot.Y += -g.YAdvance.Round()
```
hajimehoshi, line 157: Question: should Y advance earlier than drawPoints?

benoitkugler: Actually I think both ways are OK. As it is, the first glyph is draw under the horizontal line Y=0, so that it is in the image bounds. 
In your case, it is perhaps better to display the first glyph over the line.

hajimehoshi: > Actually I think both ways are OK.

YAdvance values are different among Japanese glyphs and English glyphs, so I don't understand why both could be ok.

hajimehoshi:
> Could you try and find where your pipeline differs from this [one](https://github.com/go-text/typesetting/blob/vertical-scripts/shaping/render_test.go) ?

This test code doesn't use Segmemter's Split, right? If so, I am not sure why my results is different... I am using Segmenter's Split, and without this. all the glyphs are upright.

shaping/render_test.go
benoitkugler: Well, if the advances are y1, y2, ... yn the dot is either at 
0, y1, y1+y2, y1+y2+y3, ... 
or at 
y1, y1+y2, y1+y2+y3, ...
so that the whole run is the same, but shifted by y1

shaping/render_test.go
benoitkugler: Oh, I _think_ you are correct here ! I'll double check tomorrow

shaping/render_test.go
hajimehoshi: When rendering "あaあ", and "あ" has 10 y advance and "a" has 5 y advance,

1. あ: 0, a: 10, あ: 15
2. あ: 10, a: 15, あ: 25

And in the case of 1, "a" and the last "あ" can be overlapped.

hajimehoshi:
image

Here is the current result. Thanks to your test code, I could fix some issues in my code. The last thing is Japanese texts are unnaturally shifted due to `Glyph.YOffset`.

shaping/render_test.go
whereswaldon: I do not understand why scenario 1 is a problem. That seems correct to me. You draw the glyph before shifting the dot by the advance. Is the glyph for `a` bigger than its Y advance? Could you share a screenshot of the issue?

hajimehoshi: 1. Apply this patch onto this PR
```diff
diff --git a/shaping/render_test.go b/shaping/render_test.go
index 6b056f9..e47cd3e 100644
--- a/shaping/render_test.go
+++ b/shaping/render_test.go
@@ -154,8 +154,6 @@ func drawHRun(out Output, img *image.RGBA, dot image.Point) image.Point {
 func drawVRun(out Output, img *image.RGBA, dot image.Point) image.Point {
        sizeFactor := float32(out.Size.Round()) / float32(out.Face.Upem())
        for _, g := range out.Glyphs {
-               dot.Y += -g.YAdvance.Round()
-
                minX := dot.X + g.XOffset.Round() + g.XBearing.Round()
                maxX := minX + g.Width.Round()
 
@@ -176,6 +174,8 @@ func drawVRun(out Output, img *image.RGBA, dot image.Point) image.Point {
 
                // draw the advance
                drawHLine(img, image.Pt(0, dot.Y), img.Bounds().Dx(), red)
+
+               dot.Y += -g.YAdvance.Round()
        }
 
        return dot
```

2. Run `go test ./shaping/ -run=TestShapeVerticalScripts`
3. Open `/tmp/shape_vert_mixed.png`

image

benoitkugler:
The YOffsets are indeed incorrect: this is actually a bug in harfbuzz (#126).
Since harfbuzz already position the glyphs under the origin, this will also change the implementation of "sideways" mode. I'll correct it and ping when its done.

benoitkugler:
Alright, "sideways" mode should work as well now. I've also added some helper functions `Output.{ToFontUnit, FromFontUnit}` to make drawing outlines easier.

There is one issue remaining when laying out mixed "sideways" and "upright" : the baseline are not nicely aligned. 
I think this is in the scope of go-text to properly adjust it, but I'm not sure how hard it is. 
Is is enough to use the text orientation or do we need more information ? There is a dedicated table in Opentype fonts : https://learn.microsoft.com/en-us/typography/opentype/spec/base which should be helpful. 
Also, this concerns horizontal text as well, so perhaps we should postpone this task to another PR ?

hajimehoshi:
image

Now everything seems fine. Thank you very much!

> Also, this concerns horizontal text as well, so perhaps we should postpone this task to another PR ?

I agree with you.

benoitkugler:
> Now everything seems fine except for the baseline. Thank you very much!
> 
> > Also, this concerns horizontal text as well, so perhaps we should postpone this task to another PR ?
> 
> I agree with you.

@hajimehoshi  Well, thank you very much for all the testing/debugging, complicated by the offset bug.. 

So @andydotxyz, @whereswaldon I think this PR is ready for review now :)

shaping/render_test.go
```diff
@@ -0,0 +1,185 @@
...
+		maxY := minY - g.Height.Round()
+
+		drawRect(img, image.Pt(minX, minY), image.Pt(maxX, maxY), green)
+
+		// draw the dot
+		drawPoint(img, dot, black)
+
+		// draw a sketch of the glyphs
+		glyphData := out.Face.GlyphData(g.GlyphID).(api.GlyphOutline)
+		drawGlyph(&out, img, dot.Add(image.Pt(g.XOffset.Round(), g.YOffset.Round())), glyphData, black)
```
hajimehoshi, line 145: `dotWithOffset` here?

benoitkugler:
> There is one issue remaining when laying out mixed "sideways" and "upright" : the baseline are not nicely aligned.

So, I've explored a little bit the Opentype BASE table and thought about baseline. I think there are two different concepts : 

- the general support of mixed baselines in a single line (for both horizontal and vertical directions), which is tricky and requires font knowledge 
- the adjustment of 'sideways' and 'upright' text

Implementing the first one is probably desirable, but postponed to another PR.

Implementing the second one is visually more important, and may be done quite simply, as in [582c627](https://github.com/go-text/typesetting/pull/124/commits/582c627c210bd60f46204011cb01b93418ae433a). Now, calling `Line.AdjustBaseline` just before rendering will make 'sideways' text follow the same convention than 'upright', that is roughly placing the baseline in the middle of each glyph.

We may want to fine tune the implementation of `Line.AdjustBaseline` latter, but it seems to be a good enough start. 

@hajimehoshi Could you confirm that using `Line.AdjustBaseline` visually improves the rendering ?

whereswaldon (approved):
@benoitkugler Thanks, as always, for the excellent work here. It's exciting to see us grow better vertical text support. The changes make sense, and seem to be implemented at the right level of abstraction. It's unclear to me whether this needs a second review from @andydotxyz. I don't think the public API is breaking, but it is changing (the new exported fields/methods on Direction).

@hajimehoshi Thank you so much for the excellent bug reports and features that drove this change. Thanks to your efforts, Ebitengine, Gio, Fyne, and (hopefully) many more projects will now have more correct text shaping.

opentype/api/font.go
```diff
@@ -107,6 +107,21 @@ type GlyphOutline struct {
 	Segments []Segment
 }
 
+// Sideways udpates the coordinates of the outline by applying
```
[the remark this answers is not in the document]
benoitkugler: Thank you

hajimehoshi:
@benoitkugler 
image

Now everything looks fine. Thank you very much!

What I did is to add these lines to expect the side effect of `AdjustBaseline`. I hope this is correct.

```go
line := shaping.Line{out}
line.AdjustBaseline()

// After that, `out` is directly used and `line` is no longer used.
```

hajimehoshi:
image

I found an interesting thing: the baseline varies for an English sentence. In the above figure, the existence of `g` matters. Is this intended?

hajimehoshi:
> I found an interesting thing: the baseline varies for an English sentence. In the above figure, the existence of g matters. Is this intended?

image

Unfortunately this is an issue as the baseline are not alinged. with English texts. In the above figure, I rendered `"Hello\nEbitengine\nWorld"` with the same line spaces, but the baseline are not alinged as expected.

whereswaldon:
Perhaps we should align the text not by the ascent/descent of the particular glyphs within it, but by the face's advertised ascent/descent? That should be independent of specific glyph contents, and it should look decent.

whereswaldon:
> > I found an interesting thing: the baseline varies for an English sentence. In the above figure, the existence of g matters. Is this intended?
> 
> image
> 
> Unfortunately this is an issue as the baseline are not alinged with English texts. In the above figure, I rendered `"Hello\nEbitengine\nWorld"` with the same line spaces, but the baseline are not alinged as expected.

This particular issue is more of a line-spacing/line-height problem. The typesetting repo doesn't currently implement this part of the text display process internally. It's up to applications to decide where to start the baseline of each line of text. See the discussion here for some examples of the complexities inherent in this space: https://github.com/gioui/gio/pull/123

hajimehoshi:
> Perhaps we should align the text not by the ascent/descent of the particular glyphs within it, but by the face's advertised ascent/descent? That should be independent of specific glyph contents, and it should look decent.

I agree with you. A face's metrics should be used rather than each glyph's metrics.

>  It's up to applications to decide where to start the baseline of each line of text. 

OK, but if an application moves the dot position for each line with a constant interval, each line should be aligned, right?

benoitkugler:
> OK, but if an application moves the dot position for each line with a constant interval, each line should be aligned, right?

Well, in theory, the shaper (harfbuzz) could decide to move glyphs up (for instance if the font provides a positioning table), so that I'm not sure this promise holds. 
Having said that, I agree that we should not change the baseline when it is not needed.

benoitkugler:
This is what I get using the font line extents (left) vs glyph extents (right). 
It seems to me the left version still feels a bit akward..

shape_vert_mixed

hajimehoshi:
> Also, consider lines with different text sizes, or even with different fonts.

Yeah, but in my example, the same size and face were used. I'm OK even if lines are not aligned when multiple faces or sizes are used.

hajimehoshi:
> It seems to me the left version still feels a bit akward..

In Chrome browsers, the baseline seems the same as the right. 'g' is put so that the glyph is extended to the left side. I think this makes sense.

image

whereswaldon:
> > OK, but if an application moves the dot position for each line with a constant interval, each line should be aligned, right?
> 
> Well, in theory, the shaper (harfbuzz) could decide to move glyphs up (for instance if the font provides a positioning table), so I'm not sure this promise holds. Also, consider lines with different text sizes, or even with different fonts.

I think conceptually it's correct to shift the gap between dots by a fixed value. It's true that situationally the glyphs may look misaligned if the shaper did something fancy, but that would be a feature, not a bug.

whereswaldon:
> > It seems to me the left version still feels a bit akward..
> 
> In Chrome browsers, the baseline seems the same as the right. 'g' is put so that the glyph is extended to the left side. I think this makes sense.
> 
> image

To achieve this, would we shift the glyphs by half the distance from the face's baseline to its ascent? I'd expect that to visually center the above-the-baseline portion of the text. I don't know for sure if this would be correct for all scripts, but it certainly would be for Latin and CJK.

benoitkugler:
> To achieve this, would we shift the glyphs by half the distance from the face's baseline to its ascent? I'd expect that to visually center the above-the-baseline portion of the text. I don't know for sure if this would be correct for all scripts, but it certainly would be for Latin and CJK.

Hum, I've just tried that and it seems to move the glyphs a bit too much down. I think the proper way to go would be to align the baseline : ideographic and roman.
So we would detect what baseline to use based on the script and align them.

However, it seems to me this does not solve the "only english" example. Maybe, in this case we could shift by the font advance as you suggested earlier.

benoitkugler:
Well, what do you think of handling the baseline issue in a separate PR ? This one is already large and it seems we need more discussion on how to do it anyway.

whereswaldon:
> Well, what do you think of handling the baseline issue in a separate PR ? This one is already large and it seems we need more discussion on how to do it anyway.

Fine by me

hajimehoshi:
> Well, what do you think of handling the baseline issue in a separate PR ? This one is already large and it seems we need more discussion on how to do it anyway.

+1. I'm fine with this issue addressed in another PR. I really appreciate your team's hard work!

hajimehoshi:
@andydotxyz friendly ping

di/direction.go
```diff
@@ -14,30 +18,52 @@ const (
...
-	case DirectionBTT, DirectionTTB:
+	if d.IsVertical() {
 		return Vertical
-	default:
-		return Horizontal
 	}
+	return Horizontal
 }
 
+// SwitchAwis switch from horizontal to vertical (and vice versa), preserving
```
andydotxyz, line 49: SwitchAxis probably ;) The linter should have caught this, perhaps our CI needs a few improvements.
Possibly "switches" not "switch" as well if this line is being updated.

shaping/output.go
```diff
@@ -183,40 +200,131 @@ func (o *Output) advanceSpaceAware() fixed.Int26_6 {
...
+// For vertical text, it centralizes 'sideways' runs, so
+// that text with mixed 'upright' and
+// 'sideways' orientation is better aligned.
+//
+// This is currently a no-op for horizontal text.
+//
+// Note that this method only update cross-axis metrics,
+// so that the advance is preserved. As such, it is valid
+// to call this method after line wrapping, for instance.
+func (l Line) AdjustBaseline() {
```
andydotxyz, line 296: Reading the doc implies this should be called `AlignBaselines` instead.
Just a thought, not a change request.

shaping/render_test.go
```diff
@@ -0,0 +1,195 @@
+package shaping
```
andydotxyz, line 1: We might want to look at whether this is needed in the typesetting repo as there is more render work in the `render` repository, it would be a shame to duplicate for test purposes.

andydotxyz (changes requested):
Thanks for this, and my apologies for the delay.
I had hoped to test this by using it in our render package but I am not managing to get the time for such coding sorry.

A couple of comments inline, the only one that is a change request being the typo in docs.
I do wonder though, and may welcome other comments, if adding the rasterising render code here might start duplicating the `render` repo and whether it would be worth testing this through usage of the rasteriser we maintain as a public API?

di/direction.go
benoitkugler: Updated, thank you.

shaping/output.go
benoitkugler: I've added an "s" to imply each run may be shifted.

shaping/render_test.go
benoitkugler: I agree this file seems a bit out of place, but, in the meantime, I found this quick and dirty visualisation tool very handy when developing. 

As you said, a better solution would be to use `render`. Perhaps we can keep this file for now, and once `render` supports vertical text, move the examples there and remove `shaping/render_test.go` ?

andydotxyz (approved):
Thanks

PR #125, opened 2023-12-11
Proposing review guidelines

andydotxyz:
As discussed in passing - this should help to speed up smaller changes.


benoitkugler (approved):
Nice proposal, thank you ! 

Just a question : we will update the github protection rules to reduce the number of required owner reviews to one for all PRs, right ?

whereswaldon (approved):
Works for me. We will have to change the PR policy to only strictly require 1 approval in order for the smaller changes portion to work, but I'm not worried about our ability to respect the policy.

whereswaldon:
I have updated the branch protection rule to only *require* one code owner review.

andydotxyz:
Thanks so much :) yes it will require a little care to keep the policy respected but I'm sure we can manage appropriately.

PR #126, opened 2023-12-12
[harfbuzz] Properly scale glyph X and Y offsets

benoitkugler:
Previously, the user provided Size factor was not applied to the offsets.
This is particularly visible for vertical texts (the bug was found working on #124).


whereswaldon (approved):
Oh wow, good fix!

PR #129, opened 2023-12-17
fontscan: add function to return all matched font locations

akiyosi:
I wanted to get all of the font locations that matched a given font family.
This PR adds such a function.

### DIFF
--- a/fontscan/fontmap.go
+++ b/fontscan/fontmap.go
@@ -334,6 +334,22 @@ func (fm *FontMap) FindSystemFont(family string) (Location, bool) {
 	return Location{}, false
 }
 
+// FindSystemFonts is the same as FindSystemFont, but returns all matched fonts.
+func (fm *FontMap) FindSystemFonts(family string) []Location {
+	var locations []Location
+	family = meta.NormalizeFamily(family)
+	for _, footprint := range fm.database {
+		if footprint.isUserProvided {
+			continue
+		}
+		if footprint.Family == family {
+			locations = append(locations, footprint.Location)
+		}
+	}
+
+	return locations
+}
+
 // SetQuery set the families and aspect required, influencing subsequent
 // `ResolveFace` calls.
 func (fm *FontMap) SetQuery(query Query) {


benoitkugler:
Hi @akiyosi and thank you for this contribution.

Out of curiosity, could you elaborate on why you need such function, and what is the use case you would like to support ?

akiyosi:
@benoitkugler 
Thanks for your comment.

I am creating a GUI application in Go using [Go Qt binding](https://github.com/therecipe/qt).
The issue I have is that Qt applications are very slow to start up on Windows, so I am trying to speed up the startup process by loading font information directly with Qt's low-level font interface called [QRawFont](https://doc.qt.io/qt-6/qrawfont.html#QRawFont-1).

The reason for the slow startup of the Qt application on Windows was that it took time to load a large number of installed fonts when building the font database called QFontDatabase. In addition, the process of building the QFontDatabase seemed unavoidable when using [QFont](https://doc.qt.io/qt-6/qfont.html), a common interface used by Qt applications to render text.

Further investigation revealed that the QFontDatabase construction process could be avoided by using QRawFont.
However, in order for QRawFont to load a specific font, it requires the file path of that font to be specified directly. I am wondering if `go-text/typesetting/fontscan` could be used to obtain this font file path. `FindSystemFont()` returns the first font file path that matches, so I would like to get all the font files that match and combine them to assemble the QRawFont for the Qt I need.

akiyosi:
@benoitkugler 
Yes, they are intended.

I am considering a mechanism whereby the application-side logic loads the appropriate font based on the font file filename obtained from `MatchedSystemFont()`.

fontscan/fontmap.go
```diff
@@ -334,6 +334,22 @@ func (fm *FontMap) FindSystemFont(family string) (Location, bool) {
 	return Location{}, false
 }
 
+// MatchedSystemFonts return all matched fonts
+func (fm *FontMap) MatchedSystemFonts(family string) []Location {
```
benoitkugler, line 338: Since the existing `FindSystemFont` function is very similar, I would prefer naming it `FindSystemFonts`.

fontscan/fontmap.go
```diff
@@ -334,6 +334,22 @@ func (fm *FontMap) FindSystemFont(family string) (Location, bool) {
 	return Location{}, false
 }
 
+// MatchedSystemFonts return all matched fonts
```
benoitkugler, line 337: What about 

// `FindSystemFonts` is the same as `FindSystemFont`, but returns all matched fonts.

benoitkugler (changes requested):
Good addition overall, thank you ! 
I've added two suggestions for better consistency with the existing `FindSystemFont`.

akiyosi:
@benoitkugler 
Thanks for the review! That makes sense 👍 
I fixed it.

whereswaldon:
@akiyosi Could you post a comment here stating that you understand that by contributing this code to typesetting, you are placing it into the public domain? See our [licenses](https://github.com/go-text/typesetting/blob/main/LICENSE) for the specific terms that this code would be subject to.

benoitkugler (approved):

akiyosi:
@whereswaldon 
Yes, I understand that the code I have contributed is subject to the license of this repository.

whereswaldon (approved):

Issue #113, opened 2023-11-26
fontscan doesn't have a way to detect whether the face is monotype or not


hajimehoshi:
https://pkg.go.dev/github.com/go-text/typesetting@v0.0.0-20231120180320-af78120ccb13/fontscan#FontMap.FontMetadata

FontMetadata returns a family name and an aspect, but doesn't return whether the font is monospace or not.

https://pkg.go.dev/github.com/go-text/typesetting@v0.0.0-20231120180320-af78120ccb13/opentype/api/metadata#Metadata

When a font is loaded via a loader, it is possible to know the monospace info.

Would it be possible to get this info via fontscan?

hajimehoshi:
Just out of curiosity, why doesn't [`font.Face`](https://pkg.go.dev/github.com/go-text/typesetting@v0.0.0-20231120180320-af78120ccb13/opentype/api/font#Font) have metadata info?

benoitkugler:
> Just out of curiosity, why doesn't [`font.Face`](https://pkg.go.dev/github.com/go-text/typesetting@v0.0.0-20231120180320-af78120ccb13/opentype/api/font#Font) have metadata info?

The metadata information are not required to shape/render glyphs, which is the primary job of face.Face. So it is about separation of concerns.

benoitkugler:
> https://pkg.go.dev/github.com/go-text/typesetting@v0.0.0-20231120180320-af78120ccb13/fontscan#FontMap.FontMetadata
> 
> FontMetadata returns a family name and an aspect, but doesn't return whether the font is monospace or not.
> 
> https://pkg.go.dev/github.com/go-text/typesetting@v0.0.0-20231120180320-af78120ccb13/opentype/api/metadata#Metadata
> 
> When a font is loaded via a loader, it is possible to know the monospace info.

We initially thought about adding this information into the query, but it turns out that : 

- computing if a font is monospace is rather expensive 
- using a special attribute is generally not how other systems handle monospace fonts. They rather use an appropriate _family_ name, either with the generic `monospace` value or with a precise font name.



> Would it be possible to get this info via fontscan?

So the idea is that we could, but it is not clear if we _should_. Is using the generic `monospace` family a decent solution for you ?
Said otherwise, what would be your use case that _really_ requires this property  ?

hajimehoshi:
> The metadata information are not required to shape/render glyphs, which is the primary job of face.Face. So it is about separation of concerns.

[...]

PR #130, opened 2023-12-21
Cleanup font packages structure

### LINKED ISSUE #116: [font, opentype/*] Reorganize and clean up font packages
We currently have several packages dealing with font loading, namely `font`, `opentype/api`, `opentype/api/font`, `opentype/api/metadata`. This is mainly an historical consequence of the fact that go-text used to depend on an external package to support parsing Opentype fonts.

I would like to tidy things up and remove what I think is _now_ an unnecessary API complication. While at it, we could also handle #113. To do so I propose to : 

- merge `opentype/api/metadata` into `opentype/api/font`, and add the metadata to the `Font` struct
- also merge  `opentype/api` into `opentype/api/font` : a shared `opentype/api` is only necessary if we want to support multiple font formats, but we settled on only supporting Opentype.
- finally merge `opentype/api/font` into `font`.  `font` would become the real entry point of our Opentype implementation
- as cleanup, move the remaing `opentype/{loader, tables}` packages to `font/{loader, tables}`

The changes would essentially be API modifications/renaming, but would be backward incompatible. One of the more subtle change user would have to account for is that the `font.{Font, Face}` types would become struct types, not pointer anymore (they are currently defined as alias to `*opentype/font.{Font, Face}`), requiring to add `*` everywhere they use it.

 It seems to me that this is still a good time to do such changes, before we enter a more stable phase. As suggested in #99, we could perhaps tag a version _before_ this change (say 0.1.0) and right after (say 0.2.0).

What do you all think ?

benoitkugler:
Fixes #116 

I'm afraid this PR will not be very pleasant to review, due to the numbers of files changed, but most changes are just renaming. 

As proposed, it contains the following refactors :  

- merge `opentype/api/font `and `opentype/api/metadata`. While at it, I've addressed #113 by adding two methods to `Font` : `Font.Describe() Description` and `Font.IsMonospace() bool`. We still have the efficient `Describe(ld *ot.Loader, buffer []byte) (Description, []byte)` used by `fontscan` to only fetch metadata without loading the whole font.
- merge `opentype` into `font` : the low level packages are now `font/cff/..` and `font/opentype/...` and `font` directly contains the implementation of font parsing and glyph loading

Users should now be able to mostly use `font` as entry point.

As stated in #116 , this PR does not solve an actual issue, but makes our module a lot more coherent, so that new users should have an easier time discovering and understanding its logic.

It introduces backward incompatible changes, the more subtle one being that `font.Font` and `font.Face` are now struct types, not pointers.


andydotxyz:
> It introduces backward incompatible changes, the more subtle one being that `font.Font` and `font.Face` are now struct types, not pointers.

Can you comment on why this change was made? How is the new way better, for example?
Just keen to understand as this is quite a wide hitting change.

benoitkugler:
> > It introduces backward incompatible changes, the more subtle one being that `font.Font` and `font.Face` are now struct types, not pointers.
> 
> Can you comment on why this change was made? How is the new way better, for example? Just keen to understand as this is quite a wide hitting change.

It is better in the sense it removes indirection in the packages structure. Right now, `font` is just used as a shortcut to `opentype/api/font`, and, if we glance at the repository tree, we find both `font` and `opentype`, which is quite akward. Since we now are in control of the font library, it is simpler to remove this indirection.

Said otherwise, if we would write the font library from scratch, we would rather directly write the code in the `font` package, right ?

I'm not pretending this solves serious issues. It is more about overall coherence and cleaning up things. Let's say I would be a bit annoyed if I had to justify the current structure to, say, new users or developers (again, putting the historical reason apart).

andydotxyz:
I see, I guess I didn't find the bit where "font.Face" was changed from a pointer, only the impact - can you remember which file it was?

benoitkugler:
So, in our current version, `font.Face` is defined here : https://github.com/go-text/typesetting/blob/48cc05a56658160c485cbdbe274907ec430f8025/font/font.go#L16

With this PR, `font.Face` would now be the actual struct, defined here : https://github.com/go-text/typesetting/blob/52dad1cb99a5193721a91810d78c4be9e8dfdf4d/font/font.go#L487

(Same goes for `font.Font`)

andydotxyz:
Thanks, that makes it quite plain. I guess in the future we should avoid aliasing to a different type to avoid this sort of need to break.

andydotxyz:
I can see why this is desirable (and in many parts it was discussed before) but I think we need to discuss in the channel what to do regarding a break this big - as it will probably break any code using this API.

andydotxyz (approved):
I don't have strong feelings either way, but the description is compelling

font/renderer.go
```diff
@@ -0,0 +1,433 @@
...
+	errEmptySbixTable   = errors.New("empty 'sbix' table")
+	errEmptyBitmapTable = errors.New("empty bitmap table")
+)
+
+type (
+	Segment      = ot.Segment
+	SegmentPoint = ot.SegmentPoint
+)
+
+// GlyphData describe how to graw a glyph.
```
whereswaldon, line 25:
```suggestion
// GlyphData describe how to draw a glyph.
```

whereswaldon (approved):
I noticed a few typos (that were probably present before this PR) and had an unimportant question, but this looks good!

Issue #99, opened 2023-08-15
Create releases once a new version is ready


Jacalz:
It would be good to create releases once you feel that the code is in a good shape. I find it hard to know exactly when the package should be updated for Fyne (Andrew probably knows though). Creating releases would make it easier for me, and others, to know when it makes sense to update and to what degree and update is big or small (i.e. stable for minor or patch release of Fyne). It also looks tidier, and is easier to look up versions, with versions in the go.mod.

I understand that this project is in rapid development (with breakages probably?) and as such I suggest to start with v0.x.y releases. For minor fixes y can be incremented and x for larger work.

whereswaldon:
Sure; we can certainly do that. The current API would probably be fine for a v0.1.0. We just need to figure out our process for deciding when to release a new version.

Jacalz:
This seems to have been solved a long time ago. Thanks :)

PR #132, opened 2023-12-26
Update README.md about breaking changes

benoitkugler:
As discussed on Slack, I've added a small note about breaking changes and  v0.x.y versions.

This should be tagged as v0.1.0. 

This also implies we will regularly group commits into tagged versions, fixing #99


andydotxyz (approved):
That works for me, thanks for this.

PR #134, opened 2024-01-26
[segmenter] Add word segmentation

benoitkugler:
This PR adds support for breaking text on words. The entry point is the new `WordIterator` type and the related `Segmenter.WordIterator()` method.

My personal use case is to detect words when doing hyphenation. 

For other uses cases, it might for example be useful to implement functionality such as "Select current word".

Note that we also have some (minor) changes on generated data since I've updated the Unicode version to 15.1.


whereswaldon (approved):
Nice work! Seems straightforward to me.

andydotxyz (approved):
Cool :)

Issue #140, opened 2024-03-14
harfbuzz: shaping with Unifont font panics


dominikh:
Input program:

```
package main

import (
	"log"

	"github.com/go-text/typesetting/fontscan"
	"github.com/go-text/typesetting/harfbuzz"
	"github.com/go-text/typesetting/language"
)

func main() {
	fmap := fontscan.NewFontMap(log.Default())
	if err := fmap.UseSystemFonts(""); err != nil {
		log.Fatalln("couldn't load system fonts:", err)
	}
	s := []rune("a")
	fmap.SetQuery(fontscan.Query{Families: []string{"unifont"}})
	face := fmap.ResolveFace(s[0])
	if face == nil {
		log.Fatal("couldn't find any font")
	}

	family, _ := fmap.FontMetadata(face.Font)
	if family != "unifont" {
		log.Fatalf("loaded %q instead of %q", family, "unifont")
	}

	buf := harfbuzz.NewBuffer()
	buf.Props.Language = "en-us"
	buf.Props.Script = language.Latin
	buf.Props.Direction = harfbuzz.LeftToRight
	buf.AddRunes(s, 0, 1)
	font := harfbuzz.NewFont(face)
	buf.Shape(font, nil)
}
```

Output when running:

```
2024/03/14 02:30:59 fontconfig referenced path "local.conf", but it could not be resolved to a real path
2024/03/14 02:30:59 missing fontconfig include /home/dominikh/.config/fontconfig/fonts.conf: skipping
2024/03/14 02:30:59 missing fontconfig include /home/dominikh/.fonts.conf.d: skipping
2024/03/14 02:30:59 missing fontconfig include /home/dominikh/.fonts.conf: skipping
2024/03/14 02:30:59 missing fontconfig include /nix/store/40nlkb4c2sjn588dh761yh6qbxazvlhg-home-manager-path/etc/fonts/conf.d: skipping
2024/03/14 02:30:59 missing fontconfig include /nix/store/40nlkb4c2sjn588dh761yh6qbxazvlhg-home-manager-path/etc/fonts/fonts.conf: skipping
[...]

PR #142, opened 2024-03-17
[opentype] sanitize invalid LangSys entries in GSUB/GPOS tables

benoitkugler:
It seems some fonts have corrupted LangSys entries. This commit ignores out of range feature indices. 
For #140


whereswaldon (approved):
Thank you @benoitkugler for the quick fix!

Issue #144, opened 2024-03-25
fontscan: improve caching of font collections


dominikh:
If I understand correctly, `(*FontMap).loadFont` calls `Footprint.loadFromDisk`, which might parse an entire TTC, only to return a single `font.Face`, which gets cached by `loadFont`. ISTM that if we need multiple fonts from the same font collection, then we repeat a lot of parsing. Shouldn't we instead cache all faces from the collection?

benoitkugler:
Nice catch, thank you. `loadFromDisk` should return the whole slice of faces, which should be added in the cache in `loadFont`

benoitkugler:
Hum, it is not as simple as I had thought, because the other faces in a font collection do not have the same footprint. And, for now, we do not have a simple Face -> metadata method (this is #113).

dominikh:
Well, alternatively, we shouldn't use LoadTTC and instead only load the requested font from the collection, not all fonts. We'll still parse the TTC header repeatedly, but we won't have to load every font included in the collection every time.

PR #148, opened 2024-03-26
[fontscan] only load the required face in font collections

benoitkugler:
This is for #144 , thank you @dominikh for the suggested fix. 
 
Note that we still do a little bit more than strictly needed, since `NewLoaders` read all the face _headers_. But, since parsing the header of a font is really nothing compared to parsing its tables, I think this is enough.


andydotxyz (approved):
Nice fix :)

whereswaldon (approved):
Excellent!

PR #153, opened 2024-03-29
Better case for Script

### LINKED ISSUE #152: [language] Better script casing
As an aside (which I'm happy to split into its own issue if need be), the handling of language.Script is also case-sensitive, but language.ParseScript doesn't do any normalization. So shaping with the script `Arab` doesn't work correctly, but shaping with `arab` does. That's particularly confusing because script names are preferably written in title-case.

_Originally posted by @dominikh in https://github.com/go-text/typesetting/issues/150#issuecomment-2023263124_

benoitkugler:
As @dominikh noticed, we currently use an odd convention for the binary representation of `Script`. I'm not sure what my heuristic was, but it seems simply better to adopt the general title-case convention (`Arab`, not `arab`). 

The changes has few impact on the code, since we use constants anyway. However, the cache version number of fontscan has to be bumped since the current cache stores the scripts with the lowercase convention. 

Closes #152


whereswaldon (approved):
Thank you!

hajimehoshi:
This is a good change, as Ebitengine cared about go-text's behavior (https://github.com/hajimehoshi/ebiten/blob/caecd81f4e9c2f8b3c7224956fc65538220b18aa/text/v2/gotext.go#L269-L272), and [ISO 15924 says this is case-insensitive](https://www.unicode.org/iso15924/faq.html#17).

andydotxyz (approved):

PR #154, opened 2024-03-29
shaping: ensure all-whitespace strings are assigned a face

whereswaldon:
This commit ensures that strings of only whitespace runes (and other ignorable runes for the purposes of face selection) are actually assigned a face at all. The prior code could accidentally leave the face nil for inputs of all whitespace, resulting in unexpected line metrics for the whitespace.

I'd like to cherry pick this onto v0.1.0 to create v0.1.1. That's why this changeset is based on v0.1.0 instead of main.


benoitkugler:
I've just added one test for trailing spaces (which worried me at first).

Looks excellent otherwise, thank you !

benoitkugler (approved):

PR #158, opened 2024-04-05
[font] Add a glyph extents cache

benoitkugler:
I've noticed typesetting shows somewhat poor performance when using fonts with CFF tables (usually marked with an .otf extension).

A quick benchmark revealed that the bottleneck is in the `[Face.GlyphExtents]` method, which is called (in `HarbuffShaper.Shape`) for every glyph.
Contrary to the 'glyf' table used in Truetype font files, CFF tables do not expose this metrics : we have to parse the whole glyph to compute it. 
The issue will also arise with variable fonts, for which we also have to apply the variations to the glyph description in order to compute its extents. 

Thus, I propose to add a cache for glyph extents to the `[Face]` struct. 

The following benchmark shows a net improvement for CFF tables (Raleway-v4020-Regular.otf) and variable fonts (Commissioner-VF.ttf#01). 
Performances for Truetype (DejaVuSans.ttf) or variable fonts with no variations activated (Commissioner-VF.ttf) are unchanged.

|   | old.txt | new.txt   |
| :---         |     :---:      |          ---: | 
| DejaVuSans.ttf-4            |  1.200m ± 15% |   1.145m ± 2%   -4.57% (p=0.000 n=10)  |
| Raleway-v4020-Regular.otf-4 | 11.293m ± 15% |   2.315m ± 1%  -79.50% (p=0.000 n=10)  |
| Commissioner-VF.ttf-4       |  2.412m ±  5% |   2.364m ± 5%   -2.01% (p=0.019 n=10)  |
| Commissioner-VF.ttf#01-4    | 15.231m ±  2% |   2.500m ± 2%  -83.59% (p=0.000 n=10)  |
 
The side effect of this change is that fields on `[Face] `now require setters (since updating them will usually change the glyph extents). Also, I've added a constructor (`[NewFace]`) to initialize the cache.
The other change for users is that, now, a `[Face]` should be stored and reused, whereas it was previously a cheap object which could be short-lived.

### DIFF
--- a/font/cache.go
+++ b/font/cache.go
@@ -16,6 +16,9 @@ func (ec extentsCache) get(gid GID) (GlyphExtents, bool) {
 }
 
 func (ec extentsCache) set(gid GID, extents GlyphExtents) {
+	if int(gid) >= len(ec) {
+		return
+	}
 	ec[gid].valid = true
 	ec[gid].extents = extents
 }
--- a/font/cache.go
+++ b/font/cache.go
@@ -1,0 +1,38 @@
+package font
+
+type glyphExtents struct {
+	valid   bool
+	extents GlyphExtents
+}
+
+type extentsCache []glyphExtents
+
+func (ec extentsCache) get(gid GID) (GlyphExtents, bool) {
+	if int(gid) >= len(ec) {
+		return GlyphExtents{}, false
+	}
+	ge := ec[gid]
+	return ge.extents, ge.valid
+}
+
+func (ec extentsCache) set(gid GID, extents GlyphExtents) {
+	ec[gid].valid = true
+	ec[gid].extents = extents
+}
+
+func (ec extentsCache) reset() {
+	for i := range ec {
+		ec[i] = glyphExtents{}
+	}
+}
+
+func (f *Face) GlyphExtents(glyph GID) (GlyphExtents, bool) {
+	if e, ok := f.extentsCache.get(glyph); ok {
+		return e, ok
+	}
+	e, ok := f.glyphExtentsRaw(glyph)
+	if ok {
+		f.extentsCache.set(glyph, e)
+	}
+	return e, ok
+}
--- a/font/font.go
+++ b/font/font.go
@@ -54,7 +54,7 @@ func ParseTTF(file Resource) (*Face, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &Face{Font: ft}, nil
+	return NewFace(ft), nil
 }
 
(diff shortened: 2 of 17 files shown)


font/cache.go
```diff
@@ -0,0 +1,38 @@
...
+		return GlyphExtents{}, false
+	}
+	ge := ec[gid]
+	return ge.extents, ge.valid
+}
+
+func (ec extentsCache) set(gid GID, extents GlyphExtents) {
+	ec[gid].valid = true
+	ec[gid].extents = extents
+}
```
whereswaldon, line 21: `get` has a range check to ensures that `gid` is in bounds, but `set` does not. Seems like we're risking OOB access here.

Also, I would expect that GIDs are accessed somewhat sparsely and thus we'd end up with a very sparsely-populated slice. It might be more memory efficient to use a different data structure or to store elements sorted by GID that we can resolve with a binary search (though inserting would be expensive then).

whereswaldon (commented):
I think the intent of the changeset is good, but I have some questions about the cache implementation.

font/cache.go
benoitkugler: > `set` does not. Seems like we're risking OOB access here.

The idea is that we only use `set` if `glyphExtentsRaw` returns true, which will never happen if the GID is out of bounds.

> Also, I would expect that GIDs are accessed somewhat sparsely and thus we'd end up with a very sparsely-populated slice. It might be more memory efficient to use a different data structure or to store elements sorted by GID that we can resolve with a binary search (though inserting would be expensive then).

Your right in theory. In practice, I'm not sure the additional memory usage is very significant, compared to the bytes needed to encode glyph segments, for instance.
Since it is entirely internal, perhaps such an optimization could be postponed to a subsequent PR ?

benoitkugler: I've updated `TestFontLoadHeapSize` to backup my point. It  yields the following measurements : 

arabic:amiri regular On disk: 540 KB, font memory: 2248 KB, face memory 136 KB
latin:go regular On disk: 145 KB, font memory: 344 KB, face memory 14 KB
latin:roboto regular On disk: 164 KB, font memory: 454 KB, face memory 26 KB
arabic:noto sans arabic regular On disk: 747 KB, font memory: 3377 KB, face memory 31 KB
free serif regular On disk: 1647 KB, font memory: 5588 KB, face memory 176 KB

The additional memory used by the cache is, in the worst case, around 5% of the total memory.

benoitkugler: Alright, I've added a bounds check in `set`

whereswaldon (approved):
Thank you!

PR #161, opened 2024-04-27
[opentype/tables] parse postscript names

benoitkugler:
I'm working on a font sub-setter to trim font files bundled in PDF. One heavy table is the 'post' one : it contains a (string) name for every glyph. The wire format is a byte slice where strings are Pascal encoded : [length, data, length, data, etc...] 
This format does not allow direct access at a given index, and also makes it harder to remove the unused names. 

We already process theses names in the `font.NewFont` constructor, and I would like to reuse the logic (currently hidden). To do so, this PR moves the parsing to the `opentype/tables` package, so that we can directly use the `tables.PostNames20.Strings` field.

There is no change for end users (only the consumers of the low level `tables` package will be impacted).


andydotxyz:
> I just realized that this has been in limbo for ~9 months

Oops, sorry I missed that too.

I think the work needs to be rebased to get the CI to pass (as it's been so long) can you do that @benoitkugler?

benoitkugler:
Closing in favor of #188 (sorry, Git mistakes...)

PR #162, opened 2024-04-28
[harfbuzz] reuse shaperOpentype instead of allocating a new one

benoitkugler:
This is a memory optimization : previously, the `shaper` field (with type `*shaperOpentype`) was allocated at each `harbuzz.Shape` call.

Removing the pointer and simply resetting the fields instead reduces memory usage. 

This is very similar to https://github.com/go-text/typesetting/pull/106


andydotxyz (approved):
👍

benoitkugler:
I'll merge this change, since it is an internal optimization with no API change.

PR #164, opened 2024-05-04
[shaping] Trailing spaces

benoitkugler:
Due to a missing reference operator (&), the trailing whitespaces were not collapsed properly.

An interesting side effect of this change was to discover that some tests used one `Output` several times, without copying its glyphs, leading to spurious errors when the glyphs were mutated.

### DIFF
--- a/shaping/wrapping.go
+++ b/shaping/wrapping.go
@@ -838,12 +838,13 @@ func (l *LineWrapper) postProcessLine(finalLine Line, done bool) (WrappedLine, b
 		// zero trailing whitespace advance,
 		// to be coherent with Output.advanceSpaceAware
 		if L := len(finalRun.Glyphs); L != 0 {
+			g := &finalRun.Glyphs[L-1]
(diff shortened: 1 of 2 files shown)


andydotxyz (approved):
I don't fully understand the ramification around the copy but this seems sensible.

benoitkugler:
> I don't fully understand the ramification around the copy but this seems sensible.

Some tests use the same `Output` as input value for line wrapping, mutating this input, so that the next test has an input which is not the one we intended. In pseudo code, this is : 
```go
input := Output{...}
result1 := LineWrap(input, settings1) // woups, we are actually mutating the input glyph slice
result2 := LineWrap(input, settings2) // test failed, since input has changed
```

benoitkugler:
I'll go ahead and merge, since this is a bug fix with no external API change.

whereswaldon (commented):
Thanks @benoitkugler !

PR #165, opened 2024-05-07
[shaping] add support for custom letter and word spacing

benoitkugler:
This is a revisited version of #90. 

The proposed API consists of two methods `Output.AddWordSpacing()` and `Output.AddLetterSpacing()` and a convenient function `AddSpacing` accepting a slice of `Output`s.

I've exported the `Output` methods so that it may be used by external implementation of `RunIterator`.

Implementation note : by registering the letter spacing in a new `Glyph` field, the line wrapper is able to trim edge spacing on its own, without requiring the user to "clean up" the runs after wrapping. 

What do you think ?

### DIFF
--- a/shaping/wrapping.go
+++ b/shaping/wrapping.go
@@ -221,7 +221,8 @@ func inclusiveGlyphRange(dir di.Direction, start, breakAfter int, runeToGlyph []
 
 // cutRun returns the sub-run of run containing glyphs corresponding to the provided
 // _inclusive_ rune range.
-func cutRun(run Output, mapping []glyphIndex, startRune, endRune int) Output {
+// if [trimStart] is true, the leading letter spacing is removed
+func cutRun(run Output, mapping []glyphIndex, startRune, endRune int, trimStart bool) Output {
 	// Convert the rune range of interest into an inclusive range within the
 	// current run's runes.
 	runeStart := startRune - run.Runes.Offset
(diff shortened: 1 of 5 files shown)


shaping/wrapping.go
```diff
@@ -1093,7 +1098,8 @@ func (l *LineWrapper) processBreakOption(option breakOption, config lineConfig)
 		// Reject invalid line break candidate and acquire a new one.
 		return breakInvalid, Output{}
 	}
-	candidateRun := cutRun(run, l.mapper.mapping, l.lineStartRune, option.breakAtRune)
+	isFirstInLine := len(l.scratch.alt) == 0
```
whereswaldon, line 1101: My only concern here is that we're directly observing the `alt` buffer when the rest of the logic relies upon methods that hide the specifics of the buffering. Could we instead have a `candidateLen()` method that returned this quantity? It would pair nicely with the existing `candidateAppend()`.

whereswaldon (changes requested):
Thanks @benoitkugler, this looks great! I have a very minor request.

shaping/wrapping.go
benoitkugler: Sure thing. If I follow you, the implementation of `candidateLen()` would still be `return l.scratch.alt`; but it would be more robust if we later change the buffering implementation. Am I correct ?

whereswaldon: Yeah, that's precisely what I meant. :D

benoitkugler: Done in c07830e.

whereswaldon (approved):
Thank you!

andydotxyz (approved):
Cool addition

Issue #169, opened 2024-09-09
failed to get a cap height for "M PLUS 1p" font


hajimehoshi:
I tested this with [`MPLUS1p-Regular.ttf`](https://fonts.google.com/specimen/M+PLUS+1p).

```go
package main

import (
	"fmt"
	"os"

	"github.com/go-text/typesetting/font"
	gotextopentype "github.com/go-text/typesetting/font/opentype"
	golangxopentype "golang.org/x/image/font/opentype"
)

func showGoTextCapHeight() error {
	file, err := os.Open("MPLUS1p-Regular.ttf")
	if err != nil {
		return err
	}
	defer file.Close()

	loader, err := gotextopentype.NewLoader(file)
	if err != nil {
		return err
	}

	f, err := font.NewFont(loader)
	if err != nil {
		return err
	}
	face := font.NewFace(f)
	fmt.Printf("go-text/typesettings cap height: %f\n", face.LineMetric(font.CapHeight))
	return nil
}

func showGolangXCapHeight() error {
	file, err := os.ReadFile("MPLUS1p-Regular.ttf")
	if err != nil {
		return err
	}

	f, err := golangxopentype.Parse(file)
	if err != nil {
		return err
	}
	face, err := golangxopentype.NewFace(f, &golangxopentype.FaceOptions{
		Size: 12,
		DPI:  72,
	})
	if err != nil {
		return err
	}
	m := face.Metrics()
	fmt.Printf("golang.org/x/image/font/opentype cap height: %s\n", m.CapHeight.String())
	return nil
}

func main() {
	if err := showGoTextCapHeight(); err != nil {
		fmt.Fprintln(os.Stderr, err)
		os.Exit(1)
	}
	if err := showGolangXCapHeight(); err != nil {
		fmt.Fprintln(os.Stderr, err)
		os.Exit(1)
	}
}
```

The result was:

```
go-text/typesettings cap height: 0.000000
golang.org/x/image/font/opentype cap height: -6:15
```

The dependencies are:

* github.com/go-text/typesetting v0.1.1-0.20240522210117-2c045476f496
* golang.org/x/image v0.19.0

golang.org/x/image has a fallback to get cap height for old fonts: https://github.com/golang/image/blob/c82123aa1384ef5797de16b8c67017758ebaaac6/font/sfnt/sfnt.go#L1204C16-L1223

hajimehoshi:
> golang.org/x/image/font/opentype cap height: -6:15

[...]

PR #170, opened 2024-09-25
Cap height fallback

benoitkugler:
This PR adds a fallback required to retrieve cap-height and x-height values for older fonts (having a 'OS/2'  table with version older than 2). 

(The implementation follows x/image/font and harfbuzz references.)

For #169.


whereswaldon (approved):

whereswaldon:
Since this is a straighforward bugfix, it should only need one approval. Merging.

PR #171, opened 2024-10-27
[fontmap] update Replacer for string normalization by handling null byte removal

akiyosi:
I have confirmed that the deserialized font family name contains `\x00` (e.g. MS Gothic).
This is why there is a problem with fontmap.FindSystemFonts() not matching even though the correct font name is specified.
This PR is one idea to address this problem, but where to put the normalization process may be debatable.

### DIFF
--- a/font/metadata.go
+++ b/font/metadata.go
@@ -277,7 +277,7 @@ func (fd *fontDescriptor) rawAspect() Aspect {
 	return Aspect{style, weight, stretch}
 }
 
-var rp = strings.NewReplacer(" ", "", "\t", "")
+var rp = strings.NewReplacer(" ", "", "\t", "", "\x00", "")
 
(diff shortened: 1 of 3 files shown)


akiyosi:
Hmmm, since the test is failing, would it be better to normalize the footprint.Family in FindSystemFont(). 🤔

benoitkugler:
Hi,

It seems indeed reasonnable to ignore termination bytes when comparing family names. Thank you for adressing that point.

I'm not entirely sure how to do so, I need to think about your fix a bit deeper first. 

Regarding the tests, could you rather add new ones ?

benoitkugler:
After investigation, there was actually a deeper issue. I've opened #172 and #173, superseding this PR.

akiyosi:
@benoitkugler 
Thanks for the investigation. Hopefully better improvements will be made on this issue!

PR #173, opened 2024-11-03
[font] Names decoding

benoitkugler:
Some Windows fonts use UTF-16 encoding for their family names : we were missing some of them when decoding the 'name' table.

This a fix for #172 (and should also fix #171).

I've also updated the fontscan cache version, to ensure the existing footprints will be properly updated.

@akiyosi Could you confirm this fix resolves your initial issue ? Thank you !


akiyosi:
@benoitkugler 
I have confirmed that this fix also resolves #171 issue. 
Thanks :D

whereswaldon:
Since this is a straightforward bugfix, it only needs one approval. Merging.

PR #175, opened 2024-12-16
shaping: (wrapping) make whitespace trim bidi-aware

whereswaldon:
Our previous implementation of trimming the whitespace off of the end of wrapped lines assumed that the end was always the final logical run, and that the final glyph within that run was visually last. Neither of these assumptions are true in real-world bidi text. The final logical run may be embedded wtihin the center of the visual line of text, and sometimes the final glyph is the first within a run, rather than the last.

### DIFF
--- a/shaping/output.go
+++ b/shaping/output.go
@@ -146,6 +146,11 @@ type Output struct {
 	// the output in order to render each run in a multi-font sequence in the
 	// correct font.
 	Face *font.Face
+
+	// VisualIndex is the visual position of this run within its containing line where
+	// 0 indicates the leftmost run and increasing values move to the right. This is
+	// useful for sorting the runs for drawing purposes.
+	VisualIndex int32
 }
 
 // ToFontUnit converts a metrics (typically found in [Glyph] fields)
@@ -188,7 +193,12 @@ func (o *Output) advanceSpaceAware() fixed.Int26_6 {
 	}
 
 	// adjust the last to account for spaces
-	lastG := o.Glyphs[L-1]
+	var lastG Glyph
+	if o.Direction.Progression() == di.FromTopLeft {
+		lastG = o.Glyphs[L-1]
+	} else {
+		lastG = o.Glyphs[0]
+	}
 	if o.Direction.IsVertical() {
 		if lastG.Height == 0 {
 			return o.Advance - lastG.YAdvance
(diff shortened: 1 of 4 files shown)


andydotxyz (commented):
Great catch. Should there be a test added that shows this combination of items wrapping?

whereswaldon:
@andydotxyz Absolutely there should be; I've added one.

whereswaldon:
I hit a weird case with the ordering of runs this generates; I'm going to add more tests and see if I can isolate it. The issue may be in Gio instead of typesetting.

whereswaldon:
I've fixed the visual ordering of runs, but as I've been looking into including this version of typesetting into Gio, I've realized there's an additional problem. Our `advanceSpaceAware()` implementation unconditionally subtracts the width of a trailing space from the advance of a candidate run, but that's only safe to do if the candidate run is the same text direction as the overall paragraph. Otherwise, the space that you're subtracting from the end was not likely to be displayed at the end of the line, and you're artificially shortening the run in the middle of the line.

I think the fix is simple, but making it will require some new tests. I'll try to implement those tomorrow.

andydotxyz (approved):
Nice one, thanks :)

andydotxyz:
Sorry I read the messages out of order. Will review again later if there are more changes :)

shaping/output.go
```diff
@@ -146,6 +146,11 @@ type Output struct {
 	// the output in order to render each run in a multi-font sequence in the
 	// correct font.
 	Face *font.Face
+
+	// VisualIndex is the visual position of this run within its containing line where
+	// 0 indicates the leftmost run and increasing values move to the right. This is
+	// useful for sorting the runs for drawing purposes.
+	VisualIndex int32
```
benoitkugler, line 153: Do you have a particular reason to choose `int32` over `int` ?

benoitkugler (approved):
Very nice work, thank you. 
I've added a question, and I've also took the liberty to add a minor change, simplifying the 'resolveBidi' function.

shaping/output.go
whereswaldon: I was trying to make the struct smaller. We should never need more than MAXINT32 runs in a single paragraph (2 billion runs is quite a few), so I don't think it makes sense to make the struct another 4 bytes larger on 64-bit systems by using `int`.

benoitkugler: Makes sense, thank you.

whereswaldon:
Okay, I believe this is now correct. Gio's tests fail because this space-trimming allows much better wrapping that we previously had, and the test text is actually fitting into fewer lines that it used to, but the wrapping is correct. If @benoitkugler is happy with the space aware advance tweak, I think we can merge this.

andydotxyz (approved):
Looks good yes - thanks for figuring this all out.
I really need to get the bi-di in Fyne worked on so we can be at similar levels of need for this library :).

PR #177, opened 2025-01-06
Bump to Go 1.19 as base

Jacalz:
I noticed that this project still was on Go 1.17 as base. Both Gio and Ebitengine are on 1.22 while Fyne is on 1.19 so lets move to the lowest common denominator. I also updated to the latest golang.org/x/image dependency plus made the workflow maintain itself (using latest versions of actions instead of hardcoding them) to a larger extent than before.


benoitkugler (approved):
Looks good to me, thank you !

whereswaldon (approved):
Whoa, I didn't realize Fyne had moved to 1.19! Now we can use generics!

whereswaldon:
I have updated our branch protection rules to expect the new set of CI jobs that this PR moves us to.

Jacalz:
> Whoa, I didn't realize Fyne had moved to 1.19! Now we can use generics!

Cool. It was a while ago. Sorry for having to wait. I'll make sure that we are on top of this better next time :)

PR #178, opened 2025-01-11
[fontscan] Better font resolution 

benoitkugler:
Hi,

This PR aims at providing a better font resolution, especially in the fallback case (where the family asked by the user do not support the text runes).

The main changes (inspired by fontconfig) are :  

1. Implement more precise family substitution, by splitting the added family in two parts : "strong" and "weak". The "strong" ones will be tried _before_ script fallback (see below), the "weak" ones after.  
2. Merge the fallback fonts selected by _script_ into the family fallback.
3. Support language based family substitutions : for now, this is implemented by mapping scripts to language (which is not optimal but is simpler to begin with)

This is a breaking change, since the `shaping.Fontmap` interface now requires to be script aware (via the added method `SetScript`). The two defaults implementations are updated accordingly.

As a side note, the language table is updated  (meaning the internal IDs have changed), so that the cache version is bumped.

Thank you for your feedback !


andydotxyz:
Thanks for doing this. I have a concern about another breaking change and I don't think this does have to be breaking.
Outside of the package in which this is implemented, the usages are `SplitByFontGlyphs` and `Segmenter`, both of which have the `Input` context which is what the `Script` is extracted from (requiring the breaking `SetScript` change).

There are two font map implementations and one doesn't use the script, so it seems like a big change on the API level - can we not leave it only on the maps that support it, and make it optional usage and avoid the breaking change?

benoitkugler:
> Thanks for doing this. I have a concern about another breaking change and I don't think this does have to be breaking. Outside of the package in which this is implemented, the usages are `SplitByFontGlyphs` and `Segmenter`, both of which have the `Input` context which is what the `Script` is extracted from (requiring the breaking `SetScript` change).
> 
> There are two font map implementations and one doesn't use the script, so it seems like a big change on the API level - can we not leave it only on the maps that support it, and make it optional usage and avoid the breaking change?

I think that would require to type assert if the fontmap implements the extended interface (with [SetScript]). Perhaps not a big deal. 

On the other hand, is there major pain point with the breaking change ? Assuming it is properly tagged. 
I expect many users will use the concrete [fontscan.Fontmap] implementation anyhow.

andydotxyz:
Breaking changes cause people pain, and I feel they should only be done with good reason. Any "go get -u" will break compilation unless every project using this API releases at exactly the same time.

In this example it does not seem to be necessary so it would be good to avoid.

benoitkugler:
> Breaking changes cause people pain, and I feel they should only be done with good reason. Any "go get -u" will break compilation unless every project using this API releases at exactly the same time.

Fair enough. I had overlooked the fact that go get -u does not really respect the "minor" / "patch" model we use.

fontscan/match.go
```diff
@@ -118,52 +123,92 @@ func isGenericFamily(family string) bool {
...
-	return fm.selectByFamilies(cribleBuffer, footprintBuffer)
+	return fm.selectByFamilies(cribleBuffer, bufferStrong, bufferWeak)
 }
 
-// select the fonts in the fontSet matching [crible], returning their indices.
-// footprintBuffer is used to reduce allocations.
-func (fm fontSet) selectByFamilies(crible familyCrible, footprintBuffer *scoredFootprints) []int {
-	footprintBuffer.reset(fm)
+// select the fonts in the fontSet matching [crible], returning their (sorted) indices.
+// [buffer1] and [buffer2] are used to reduce allocations.
```
whereswaldon: Nit: Outdated parameter names

fontscan/fontmap.go
```diff
@@ -506,18 +506,17 @@ func (fm *FontMap) resolveForLang(candidates []int, lang LangID) *font.Face {
...
 //
-// Fonts are tried in 3 steps :
+// Fonts are tried with the following steps :
 //
 //	1 - Only fonts matching exacly one of the [Query.Families] are considered; the list
 //		is prunned to keep the best match with [Query.Aspect]
 //	2 - Fallback fonts are considered, that is fonts with similar families and fonts
-//		supporting the current script; the list is also prunned according to [Query.Aspect]
-//	3 - All fonts matching the current script (set by [FontMap.SetScript]) are tried,
+//		supporting the current script; the list is also prunned according to [Query.Aspect]4
```
whereswaldon, line 514: Nit: Trailing 4 is probably a typo.

whereswaldon (approved):
I found a couple of very minor documentation issues, but the changeset looks good to me on the whole.

fontscan/fontmap.go
benoitkugler: Thanks, fixed.

fontscan/match.go
benoitkugler: I think this was on a previous version of the changes, and has been fixed.

andydotxyz (approved):
Great change, thanks for making it backward compatible!

PR #179, opened 2025-01-11
Add back Windows and macOS to tests

Jacalz:
These were accidentally removed in my Go 1.19 update PR due to copy paste error from go-text/render which only ran on Ubuntu even before my changes there. Sorry.


andydotxyz (approved):
Thanks for putting that back :)

PR #180, opened 2025-01-15
Add language resolution

benoitkugler:
This PR adds language resolution during segmentation, enabling Harfbuzz to activate some language specific features.

The changes are roughly divided into two steps : 
- (part of) the language table is moved into the `language` package. This is a bit unfortunate but seems required to use it in the `shaping` package without import cycle. The change is (as far as I know) backward compatible, using type alias and forwarding functions.
- The `shaping.Segmenter.Split` method now includes a language resolution step. 

(You may look at the `TestShapingLanguage` for a concrete use case where it matters.)


language/language.go
```diff
@@ -156,3 +156,171 @@ func DefaultLanguage() Language {
...
+	if i, ok := binarySearchLang(l, languagesInfos[:knownLangsCount]); ok {
+		return LangID(i), true
+	}
+	if i, ok := binarySearchLang(l, languagesInfos[knownLangsCount:]); ok {
+		return knownLangsCount + LangID(i), true
+	}
+	return 0, false
+}
+
+func binarySearchLang(l Language, records []languageInfo) (int, bool) {
```
whereswaldon, line 181: Could we use a battle-hardened binary search? I don't love re-writing this algorithm if we don't have to. I think [this](https://pkg.go.dev/slices#BinarySearchFunc) should be available in Go 1.19. If not, `golang.org/x/exp/slices` should be.

whereswaldon (commented):
On the whole, this looks great! Thank you for the feature. However, I'm not sure adding another binarysearch implementation makes sense.

language/language.go
benoitkugler: `slices.BinarySearch` is only available in Go 1.21. Could we wait until we support it  ? Also, the implementation is really just a copy of what we had previously.

Jacalz: Maybe https://pkg.go.dev/sort#Search could be used until Go 1.21 is the minimum version?

benoitkugler:
The last commit follows @Jacalz suggestion and uses `sort.Search` until we support go 1.21.

andydotxyz (approved):
Nice work, looks good as far as I can see!

whereswaldon (approved):
Thanks @Jacalz for suggesting the other Search implementation! And thanks @benoitkugler for implementing the change. Looks great to me.

PR #183, opened 2025-02-18
shaping: treat truncator as proper run for bidi and rune counting

whereswaldon:
This commit fixes an oversight in our handling of text truncators. They were inserted into the text *after* bidi run reordering had already occurred, which would place them in the wrong visual position sometimes. Additionally, their rune data reflected the runes of the truncator string instead of the runes of text the truncator represents, but it is clearly more desirable for the truncator to stand in for all of the text it replaces.

This commit does change the semantics of a truncator run's Range field in the output text. It will now describe the replaced range of text. This does greatly simplify consuming the shaped text data, as you can see by how it simplified the tests. We don't have to factor out the truncator or do extra math to make sure all runes are accounted for. Instead, the sum of the rune counts of all runs will always equal the full rune count of the input text.

### DIFF
--- a/shaping/wrapping.go
+++ b/shaping/wrapping.go
@@ -973,10 +973,10 @@ func (l *LineWrapper) postProcessLine(finalLine Line, done bool) (WrappedLine, b
 // is non-zero, the final line of text returned by successive calls to WrapNextLine
 // may be truncated. The quantity of runes truncated by wrapping the line is
 // returned in [WrappedLine].Truncated. If this is non-zero AND the LineWrapper's
-// [WrapConfig].Truncator is set, the final [Output] of the truncated line will be a copy
-// of the Truncator with its Runes.Count set to the quantity of runes truncated during
-// line wrapping. The sum of all [Output].Runes.Count in the line will the number of runes
-// provided as input to the line wrapping process.
+// [WrapConfig].Truncator is set, the final line will end with an extra [Output].
+// [Output]s before the final [Output] will represent the input runes that are still
+// visible before truncation, and the final [Output] will be a copy of the Truncator
+// with its Runes.Count set to the quantity of runes truncated during line wrapping.
 func (l *LineWrapper) WrapNextLine(maxWidth int) (out WrappedLine, done bool) {
 	// If we've already finished the paragraph, don't do any more work.
 	if !l.more {
--- a/shaping/wrapping.go
+++ b/shaping/wrapping.go
@@ -973,9 +973,10 @@ func (l *LineWrapper) postProcessLine(finalLine Line, done bool) (WrappedLine, b
 // is non-zero, the final line of text returned by successive calls to WrapNextLine
 // may be truncated. The quantity of runes truncated by wrapping the line is
 // returned in [WrappedLine].Truncated. If this is non-zero AND the LineWrapper's
-// [WrapConfig].Truncator is set, the final run of the truncated will be a copy
+// [WrapConfig].Truncator is set, the final [Output] of the truncated line will be a copy
 // of the Truncator with its Runes.Count set to the quantity of runes truncated during
-// line wrapping.
+// line wrapping. The sum of all [Output].Runes.Count in the line will the number of runes
+// provided as input to the line wrapping process.
 func (l *LineWrapper) WrapNextLine(maxWidth int) (out WrappedLine, done bool) {
 	// If we've already finished the paragraph, don't do any more work.
 	if !l.more {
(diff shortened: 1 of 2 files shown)


benoitkugler:
Looks good to me, thank you !

(I've opened #184 to resolve staticcheck failure, perhaps we should wait for the CLI to be back online ?)

whereswaldon:
Rebased atop `main` to force it to pick up the staticcheck changes.

benoitkugler (approved):

shaping/wrapping_test.go
```diff
@@ -2131,6 +2133,13 @@ func TestWrappingTruncationEdgeCases(t *testing.T) {
 			lastLine := lines[len(lines)-1]
 			lastRun := lastLine[len(lastLine)-1]
 			shouldTruncate := (tc.expectedTruncated > 0 || tc.forceTruncation)
+			// The VisualIndex and rune counts on the truncator will be populated during wrapping, and thus
```
andydotxyz, line 2136: This seems confusing - we have tests documenting that data returned from the API is not to be used...
But nowhere in the API is this change explained.

whereswaldon: This comment is actually obsolete. I wrote it before realizing that we could still easily test these outputs for correctness. The two statements below modifying `trunc` populate its VisualIndex and Runes with correct expectations.

whereswaldon: I've pushed a change to drop this comment

andydotxyz (approved):
Thanks, upon re-reading the documentation I see that it fits and it was assumptions in Fyne code that did not match the documented outcomes.
It now aligns and I have updated our code so this change is non-breaking after all :)

shaping/wrapping.go
```diff
@@ -963,6 +968,14 @@ func (l *LineWrapper) postProcessLine(finalLine Line, done bool) (WrappedLine, b
 //
 // The returned line is only valid until the next call to
 // [*LineWrapper.Prepare] or [*LineWrapper.WrapParagraph].
+//
+// If the LineWrapper's [WrapConfig].TruncateAfterLines
+// is non-zero, the final line of text returned by successive calls to WrapNextLine
+// may be truncated. The quantity of runes truncated by wrapping the line is
+// returned in [WrappedLine].Truncated. If this is non-zero AND the LineWrapper's
+// [WrapConfig].Truncator is set, the final run of the truncated will will be a copy
+// of the Truncator with its Runes.Count set to the quantity of runes truncated during
```
andydotxyz, line 977: Almost, I still trip over "truncated" because `Runes.Count` is not impacted by the truncation/wrap - could "runes truncated" be "runes in the current run"?

shaping/wrapping.go
```diff
@@ -963,6 +968,14 @@ func (l *LineWrapper) postProcessLine(finalLine Line, done bool) (WrappedLine, b
 //
 // The returned line is only valid until the next call to
 // [*LineWrapper.Prepare] or [*LineWrapper.WrapParagraph].
+//
+// If the LineWrapper's [WrapConfig].TruncateAfterLines
+// is non-zero, the final line of text returned by successive calls to WrapNextLine
+// may be truncated. The quantity of runes truncated by wrapping the line is
+// returned in [WrappedLine].Truncated. If this is non-zero AND the LineWrapper's
+// [WrapConfig].Truncator is set, the final run of the truncated will will be a copy
```
andydotxyz, line 976:
```suggestion
// [WrapConfig].Truncator is set, the final run of the truncated will be a copy
```

shaping/wrapping.go
whereswaldon: I've pushed what I hope is an additional clarification. Does it help?

andydotxyz: It does, but I still struggle a little. I see the words are all in here but the implication is somewhat unclear. What if 

> If this is non-zero AND the LineWrapper's [WrapConfig].Truncator is set, the final [Output] of the truncated line will be a copy of the Truncator with its Runes.Count set to the quantity of runes truncated during line wrapping.

Becomes:

> If this is non-zero AND the LIneWrapper's [WrapConfig].Truncator is set, the final line will be split into two [Output]s. The penultimate will represent the input runes that are still visible before truncation, and the final line will be a copy of the Truncator with its Runes.Count set to the quantity of runes truncated during line wrapping.

whereswaldon: Since any line can contain any number of input runs, we have to be careful about the wording, but I see what you're going for.

> If this is non-zero AND the LineWrapper's [WrapConfig].Truncator is set, the final line will end with an extra [Output]s. [Output]s before the final [Output] will represent the input runes that are still visible before truncation, and the final [Output] will be a copy of the Truncator with its Runes.Count set to the quantity of runes truncated during line wrapping.

I think this is accurate, while preserving your point.

andydotxyz: Perfect thanks :)

andydotxyz (approved):
Now all is clear thanks!

PR #186, opened 2025-03-25
two nil guards that we need

rcoreilly:
these allow:

1. nonstandard output wrapping elements with no glyphs. we use this for math rendering, which creates a "dummy" output element for line wrapping purposes.

2. fonts without glyf table (metrics only). we use these for rendering on web platforms that use html canvas measure text but do not require glyf tables, so the font files can be smaller.

### DIFF
--- a/font/opentype/tables/glyphs_glyf_src.go
+++ b/font/opentype/tables/glyphs_glyf_src.go
@@ -39,9 +39,6 @@ type Glyf []Glyph
 // ParseGlyf parses the 'glyf' table.
 // locaOffsets has length numGlyphs + 1, and is returned by ParseLoca
 func ParseGlyf(src []byte, locaOffsets []uint32) (Glyf, error) {
-	if len(src) == 0 {
-		return Glyf{}, nil
-	}
 	out := make(Glyf, len(locaOffsets)-1)
 	var err error
 	for i := range out {
--- a/font/opentype/tables/glyphs_glyf_src.go
+++ b/font/opentype/tables/glyphs_glyf_src.go
@@ -39,6 +39,9 @@ type Glyf []Glyph
 // ParseGlyf parses the 'glyf' table.
 // locaOffsets has length numGlyphs + 1, and is returned by ParseLoca
 func ParseGlyf(src []byte, locaOffsets []uint32) (Glyf, error) {
+	if len(src) == 0 {
+		return Glyf{}, nil
+	}
 	out := make(Glyf, len(locaOffsets)-1)
 	var err error
 	for i := range out {
--- a/shaping/wrapping.go
+++ b/shaping/wrapping.go
@@ -266,6 +266,9 @@ func (option breakOption) isValid(runeToGlyph []int, out Output) bool {
 		// Check if this break is valid.
 		gIdx := runeToGlyph[breakAfter]
 		g2Idx := runeToGlyph[nextRune]
+		if gIdx >= len(out.Glyphs) || g2Idx >= len(out.Glyphs) {
+			return false
+		}
 		cIdx := out.Glyphs[gIdx].ClusterIndex
 		c2Idx := out.Glyphs[g2Idx].ClusterIndex
 		if cIdx == c2Idx {


whereswaldon:
> nonstandard output wrapping elements with no glyphs. we use this for math rendering, which creates a "dummy" output element for line wrapping purposes.

Could you provide a code sample of this use-case? I don't currently grasp what you mean, and seeing it in more context would help. Alternatively, some marked-up screenshots might be helpful.

> fonts without glyf table (metrics only). we use these for rendering on web platforms that use html canvas measure text but do not require glyf tables, so the font files can be smaller.

@benoitkugler I don't know enough to comment on what risks (if any) are imposed by deleting this check.

benoitkugler:
Regarding the 'glyf' table, I would rather return a slice with the correct length but with empty glyphs, and move the check in the `NewFont` constructor. As far as I understand, that would still fix the issue.

rcoreilly:
sorry I realized that I actually managed to avoid the glyf table issue by also omitting the loca table as well, but I had forgotten about that when I finally got around to doing this PR.

As for the dummy output, it is literally just an empty output placeholder:

```Go
		if sty.Special == rich.Math {
			o := shaping.Output{}
			o.Runes.Offset = start
			o.Runes.Count = end - start
			o.Advance = math32.ToFixed(bb.Size().X)
			sh.outBuff = append(sh.outBuff, o)
			si++ // skip the end special
			continue
		}
```

This might also be a good time to mention that the line wrapping code looks like it is difficult to extend, e.g., for this case of including other non-text elements as placeholders. There is so much complexity going on in this code that it would be rather daunting to try to replicate it, and yet it looks like there are a lot of interdependent, unexported bits to it, so it really isn't possible as far as I can tell to just intervene in one part of the logic. I haven't really had a chance to analyze it carefully however, so if I'm missing something and it actually would be possible to achieve this outcome in some other cleaner way, I'd be grateful to hear of it. Thanks!

rcoreilly:
and this is a test example of the resulting output:

image

it is using the star-tex.org tex -> dvi framework (pure Go) and the tdewolff/canvas dvi rendering code (adapted to Cogent Core) to render the math as glyph paths, that are inserted in the right place in the shaped output.

benoitkugler:
> sorry I realized that I actually managed to avoid the glyf table issue by also omitting the loca table as well, but I had forgotten about that when I finally got around to doing this PR.
> 
> As for the dummy output, it is literally just an empty output placeholder:
> 
> ```go
> 		if sty.Special == rich.Math {
> 			o := shaping.Output{}
> 			o.Runes.Offset = start
> 			o.Runes.Count = end - start
> 			o.Advance = math32.ToFixed(bb.Size().X)
> 			sh.outBuff = append(sh.outBuff, o)
> 			si++ // skip the end special
> 			continue
> 		}
> ```
> 

Not sure if that helps, but perhaps you could use the concept of cluster ; map all your runes to a single, empty glyph with the advance you need.

whereswaldon:
> sorry I realized that I actually managed to avoid the glyf table issue by also omitting the loca table as well, but I had forgotten about that when I finally got around to doing this PR.
> 
> As for the dummy output, it is literally just an empty output placeholder:
> 
> ```go
> 		if sty.Special == rich.Math {
> 			o := shaping.Output{}
> 			o.Runes.Offset = start
> 			o.Runes.Count = end - start
> 			o.Advance = math32.ToFixed(bb.Size().X)
> 			sh.outBuff = append(sh.outBuff, o)
> 			si++ // skip the end special
> 			continue
> 		}
> ```

Thank you for the sample! It clarified what you mean. Instead of a completely empty section of text, perhaps insert an Object Replacement Character as mentioned below? The real challenge here is that Shaped text is built on the assumption that underlying text data exists. Having an Output that corresponds to nothing within the text breaks invariants, but having one that corresponds to a codepoint that means "replace me" should be valid.

> This might also be a good time to mention that the line wrapping code looks like it is difficult to extend, e.g., for this case of including other non-text elements as placeholders. There is so much complexity going on in this code that it would be rather daunting to try to replicate it, and yet it looks like there are a lot of interdependent, unexported bits to it, so it really isn't possible as far as I can tell to just intervene in one part of the logic. I haven't really had a chance to analyze it carefully however, so if I'm missing something and it actually would be possible to achieve this outcome in some other cleaner way, I'd be grateful to hear of it. Thanks!

I'm the author of that particular eldritch horror portion of the code. :sweat_smile:  I think adding support for text objects wouldn't be as bad as you think, but I have the advantage of already knowing how things fit together. Roughly, I think you'd need to:

- Teach the text segmentation process to split out `U+FFFC` (the Object Replacement Character) as its own `Input`. You might use a custom segmentation process anyway, in which case you could do this yourself.
- Make the shaper detect such an "inline text object" and skip shaping it (it has no textual representation), producing a valid `Output` with no glyphs (but covering one rune of text) and an advance provided by the caller as the width of the thing the Object codepoint represents. The commit currently on this PR is part of this solution, as it enables us to process zero-glyph Outputs, but there may be some additional places where we assume non-zero glyph counts to re-examine.

The line wrapping algorithm will not break within the text object because it's a single codepoint, and it only generates breaking candidates between codepoints (really grapheme clusters). If the Output has a properly-configured advance, it might "just work" :tm:. I could be forgetting some detail, but I certainly think the changes within line wrapping itself are trivial-to-none. It's all pre-processing to feed an accurate representation of what you want into the algorithm with some possible minor tweaks to allow for a zero-glyph Output flowing through properly.

andydotxyz (approved):
The conversation is good, though this does look like a good fix anyway.

rcoreilly:
@whereswaldon thanks for all the insight into the issues here. I got all my math processing code working well today, and can confirm that this PR is all I need to make it work in my use-case, so it would be great to get this merged. Thanks!

whereswaldon (approved):
Since this change only makes us more robust, I think it is worth having. Doubly so given that it unblocks a use-case for Cogent Core.

PR #187, opened 2025-03-31
shaping(wrapping): honor newline literals and other mandatory breaks

### LINKED ISSUE #185: newline does not automatically wrap?
It does not appear that `\n` in a string causes a  line wrap in `shaping.WrapParagraph`. I could not find any discussion if this issue in the sources or issues. Is it intentional or perhaps an oversight? I can just make an outer loop to process it manually but if it might be something that should be in the shaper code, that would be good to know.

WrapParagraph does connote a string without any newlines, but on the other hand in HTML the distinction between `<br>` and `<p>` is relevant in that a `<br>` drives a line wrap but not a new paragraph. This is the context in which the issue is arising for me.

whereswaldon:
This commit updates our algorithm to respect literal newlines and other codepoints indicating a mandatory line break. Unfortunately, doing so required removing some fast-paths through the code because they worked by assuming that small quantities of text could skip line wrapping entirely.

Now that the contents of the text can force us to engage the line wrapping process in spite of whether the text would fit without wrapping, we can't skip the line wrapping process as cheaply. I've modified the most common codepath that processes single text runs so that it scans the text for mandatory breaks and is able to take the fast path if none are found, but this is more expensive than the old checks. Additionally, I've removed a second fast path that was designed to skip the normal wrapping algorithm if only one run remained to be wrapped. It was too expensive to properly scan the remaining text for mandatory breaks in a way that didn't interfere with the line wrapping algorithm.

Fixes #185

@rcoreilly Could you test this change against your use-cases for literal line breaks?

### DIFF
--- a/shaping/wrapping.go
+++ b/shaping/wrapping.go
@@ -255,6 +255,8 @@ func cutRun(run Output, mapping []glyphIndex, startRune, endRune int, trimStart 
 type breakOption struct {
 	// breakAtRune is the index at which it is safe to break.
 	breakAtRune int
+	// required indicates that the break option is mandatory.
+	required bool
 }
 
 // isValid returns whether a given option violates shaping rules (like breaking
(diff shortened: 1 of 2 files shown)


rcoreilly:
@whereswaldon this works perfectly in my tests -- very much appreciated!! :)  Hopefully the loss of the fast paths doesn't add too much time overall.

benoitkugler (approved):
Looks good to me, simple and clear. Nice work !

andydotxyz (approved):
I've not integrated this into my workflow but the code & approach looks good.

Issue #193, opened 2025-05-14
Burmese Text Rendering Issue with Noto Sans Myanmar


eroberts20:
### Description:
I'm encountering an issue where certain Burmese strings don't render correctly in go-text/typesetting with Noto Sans Myanmar, though they display perfectly in other programs and with hb-shape. It seems related to glyph reordering or positioning, especially for sequences with pre-base vowels.

### Affected Strings:
ရော့ (U+101B U+1031 U+102C U+1037)
ရော့စ် (U+101B U+1031 U+102C U+1037 U+1005)
ရော့ဘ် (U+101B U+1031 U+102C U+1037 U+1018)
ရောင်း (U+101B U+1031 U+102C U+1004 U+103A)

Commonality: All begin with U+101B U+1031 U+102C (consonant "ra" + vowel E + vowel AA), suggesting a reordering issue with U+1031.

### Steps to Reproduce:

1. Use go-text/typesetting (v0.2.0) with Noto Sans Myanmar.
2. Shape text with HarfBuzzShaper:

```
package main

import (
    "fmt"
    "os"
    "github.com/go-text/typesetting/font"
    "github.com/go-text/typesetting/font/opentype"
    "github.com/go-text/typesetting/shaping"
    "golang.org/x/image/font/sfnt"
    "golang.org/x/text/language"
)

func main() {
    fontData, _ := os.ReadFile("NotoSansMyanmar-Regular.ttf")
    fnt, _ := font.ParseTTF(fontData)
    shaper := shaping.HarfBuzzShaper{Fonts: []*sfnt.Font{fnt}}
    input := shaping.Input{
        Text:      []rune("ရော့"),
        Script:    language.MustParseScript("Mymr"),
        Language:  language.MustParse("my"),
        Direction: shaping.LeftToRight,
        Features: []shaping.FontFeature{
            {Tag: opentype.NewTag('c', 'c', 'm', 'p'), Value: 1},
            {Tag: opentype.NewTag('l', 'o', 'c', 'l'), Value: 1},
            {Tag: opentype.NewTag('m', 'a', 'r', 'k'), Value: 1},
            {Tag: opentype.NewTag('m', 'k', 'm', 'k'), Value: 1},
            {Tag: opentype.NewTag('a', 'b', 'v', 'f'), Value: 1},
            {Tag: opentype.NewTag('b', 'l', 'w', 'f'), Value: 1},
[...]

PR #195, opened 2025-05-27
[harfbuzz] fix myanmar (for #193)

benoitkugler:
This is a bug fix in the Myanmar shaper (with a new test case).


whereswaldon (approved):
Subtle! Thank you for tackling this.

PR #197, opened 2025-06-24
Add Thread Safety to Face Struct in font Package

cwarden:
Add mutex protection to the Face struct to prevent race conditions when accessing cached glyph extents and font metrics from multiple goroutines.

This ensures Face structs can be safely accessed from multiple goroutines for read operations, and ensures synchronization for writes.

Fixes race conditions detected when using go test -race.

### DIFF
--- a/font/cache.go
+++ b/font/cache.go
@@ -30,12 +30,18 @@ func (ec extentsCache) reset() {
 }
 
 func (f *Face) GlyphExtents(glyph GID) (GlyphExtents, bool) {
+	f.cacheMu.RLock()
 	if e, ok := f.extentsCache.get(glyph); ok {
+		f.cacheMu.RUnlock()
 		return e, ok
 	}
+	f.cacheMu.RUnlock()
+
 	e, ok := f.glyphExtentsRaw(glyph)
 	if ok {
+		f.cacheMu.Lock()
 		f.extentsCache.set(glyph, e)
+		f.cacheMu.Unlock()
 	}
 	return e, ok
 }
--- a/font/cache_race_test.go
+++ b/font/cache_race_test.go
@@ -1,0 +1,345 @@
+package font
+
+import (
+	"bytes"
+	"sync"
+	"testing"
+
+	td "github.com/go-text/typesetting-utils/opentype"
+	"github.com/go-text/typesetting/font/opentype/tables"
+)
+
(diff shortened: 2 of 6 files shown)


benoitkugler:
Hi,

Thank you for the PR. I'm really not sure however that we want to do that : the Face object is explicitly not thread safe, and adding synchronization complexifies the implementation quite a bit. Also, there are natural use cases where we don't require synchronization, so that adding it just decrease performance for no reason.

What is your use case ?

cwarden:
> Hi,
> 
> Thank you for the PR. I'm really not sure however that we want to do that : the Face object is explicitly not thread safe, and adding synchronization complexifies the implementation quite a bit. Also, there are natural use cases where we don't require synchronization, so that adding it just decrease performance for no reason.
> 
> What is your use case ?

I'm building an experimental library on top of Fyne and running into race conditions in typesetting, for example when multiple goroutines try to measure the length of the string.

```
WARNING: DATA RACE
Read at 0x00c00052a35c by goroutine 38:
  github.com/go-text/typesetting/font.extentsCache.get()
      /home/cwarden/go/pkg/mod/github.com/go-text/typesetting@v0.2.1/font/cache.go:14 +0x8e
  github.com/go-text/typesetting/font.(*Face).GlyphExtents()
      /home/cwarden/go/pkg/mod/github.com/go-text/typesetting@v0.2.1/font/cache.go:33 +0x28
  github.com/go-text/typesetting/harfbuzz.(*Font).GlyphExtents()
      /home/cwarden/go/pkg/mod/github.com/go-text/typesetting@v0.2.1/harfbuzz/fonts.go:137 +0x44
  github.com/go-text/typesetting/shaping.(*HarfbuzzShaper).Shape()
      /home/cwarden/go/pkg/mod/github.com/go-text/typesetting@v0.2.1/shaping/shaping.go:124 +0xdb6
  fyne.io/fyne/v2/internal/painter.shapeCallback()
      /home/cwarden/git/fyne/internal/painter/font.go:312 +0x9c
  fyne.io/fyne/v2/internal/painter.walkString()
      /home/cwarden/git/fyne/internal/painter/font.go:303 +0x424
  fyne.io/fyne/v2/internal/painter.MeasureString()
      /home/cwarden/git/fyne/internal/painter/font.go:216 +0xb8
...
```

PR #198, opened 2025-07-05
Support compiling for TinyGo

andydotxyz:
The issue was that os.UserCacheDir is missing.


fontscan/fontmap_cache_android.go
```diff
@@ -0,0 +1,10 @@
+//go:build android
```
hajimehoshi, line 1: I don't think we need this comment as the file name already includes `_android` suffix.

andydotxyz: That is true it is not needed... just a habit of mine sorry

benoitkugler:
Looks good to me, thank you for this addition !

benoitkugler (approved):

PR #200, opened 2025-09-28
Adds new methods to resolve all the fonts instead of just the first match.

maa3x:
Added functionality to retrieve the full list of supported fonts instead of just the first match. 

This could be useful for others as well.


andydotxyz:
> the ResolveForString searches for the fonts that support all the runes inside given string

Thanks for this, perhaps that docs can be clearer on that as the `string` type could be interpreted many ways - particularly when "query" is mentioned in the doc.

PR #201, opened 2025-09-29
[shaping] expose ComputeBidiOrdering

benoitkugler:
This PR expose the logic used to compute visual ordering for mixed LRT/RTL text. 

It could be used to fix go-text/render#23 in the `render` package.

It also includes a very small change so that it could also be used for vertical text (comparing `Progression` instead of `Direction`).

### DIFF
--- a/shaping/wrapping.go
+++ b/shaping/wrapping.go
@@ -889,16 +889,20 @@ func swapVisualOrder(subline Line) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


andydotxyz (approved):
Thanks - I am glad this could help renderer, I hope to get back to it after the next Fyne release is done.

I guess we need to figure out the CI issues before this can land?

benoitkugler:
Closed in favor of https://github.com/go-text/render/pull/24

PR #203, opened 2025-10-02
Refactor Unicode General Category lookup

### LINKED ISSUE #202: Go v1.25 seems to break some tests
It seems some changes to the `unicode` standard package break some of our tests, and more importantly, some Unicode Line Break rules conformity tests.

benoitkugler:
Recent changes in the standard `unicode` package have broken some tests.

This PR aims at fixing this behavior and ensuring no future change would break our tests by not relying on the standard package anymore.

There is a backward incompatible change : simplifying `LookupType` to `IsAssigned`, but it is in the `unicodedata` package which is mainly internal, so that (with proper minor version bump) it should be acceptable. 

Fixes #202


andydotxyz (changes requested):
This looks like a breaking change - any code using `unicode.LookupType` will no longer compile.
Any reason we can't leave the old API as well so it doesn't break peoples apps?

PR #205, opened 2025-12-08
Expose trimed whitespace

benoitkugler:
When using `go-text` to wrap text ending with spaces, I want to put the final space on the first line _when_ it fits (speaking about the advance). This requirement is somewhere between `DisableTrailingWhitespaceTrim=true` and `DisableTrailingWhitespaceTrim=false` : disabling trimming will miss wrapping opportunities, enabling it will always remove the space advance even when it could fit.

The workaround I've found is to to enable trimming, but to check after if the advance _including_ the space would have fit. 

The cleanest way to do so seems to expose the actual advance that was trimmed: that is implemented in this PR.

### DIFF
--- a/fontscan/fontmap_cache_android.go
+++ b/fontscan/fontmap_cache_android.go
@@ -1,5 +1,7 @@
 package fontscan
 
+import "fmt"
+
 func platformCacheDir() (string, error) {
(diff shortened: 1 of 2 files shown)


andydotxyz (approved):
Great :)

Looks like there is a conflict with other items recently merged :(

Issue #174, opened 2024-11-18
Support setting tabs width


benoitkugler:
I would like to support setting the width of tabs in a run, following the CSS '[tab-size](https://www.w3.org/TR/css-text-3/#tab-size-property)' property. 

That could be exposed with the method
```go
// SetTabsWidth alters the run, replacing the width of every tab character (U+0009).
//
// [text] is the input slice used to create the run.
func (*Output) SetTabsWidth(text []rune, spacing fixed.Int26_6)
```

(At a first glance, the implementation would be somewhat similar to `Output.AddWordSpacing`)

What do you think ?

andydotxyz:
A great idea, however I don't think it is quite that straight forward - I think you also need to have a starting position, because tab alignment is actually relative to the page and not to the first character of a run...

benoitkugler:
Well, I think what I've in mind is indeed a lot simpler than tabs alignement. I think what you describe is something about "dynamic" adjustment, depending on the position of the run with respect to the page start ?

andydotxyz:
What you outline I think only works when the tabs are right at the beginning of a line. But for example "\t\tpoint a:\t detail here" won't work as expected.

A while back I hoped to contribute the algorithm we use in Fyne for this work but licensing doesn't allow.

hajimehoshi:
Tabs work in much more complicated way in e.g. MS Word's tab stops. Are tabs really this library's scope?

benoitkugler:
> What you outline I think only works when the tabs are right at the beginning of a line. But for example "\t\tpoint a:\t detail here" won't work as expected.
> 
> A while back I hoped to contribute the algorithm we use in Fyne for this work but licensing doesn't allow.

Aah ! Indeed, you are right. Browser rendering for the string 
`\t\tpoint a:\tdetail here`
with `tab-size: 10` (meaning a tab is 10 spaces) :
![image](https://github.com/user-attachments/assets/6f872fbb-2529-4a47-a3c6-b098c616c96a)
The expected result is to align the content after a tab to a "tab grid" (displayed in red in the example). 

> Tabs work in much more complicated way in e.g. MS Word's tab stops. Are tabs really this library's scope?

Well, since it seems not trivial, I could be great to provide it for all go-text consumers right ? (If we find an API that suits everybody.)

hajimehoshi:
I'm not sure it is really possible to provide a generic model for various use cases including MS Word, but I keep my fingers crossed :)

PR #207, opened 2025-12-12
[shaping] initial support for tab alignement

benoitkugler:
This PR is a first step towards support for tabs (for #174).

The only API added is the `Line.AlignTabs` method which aligns all tabs of a given line. 

This seems enough if you only wrap text one line at a time (typically using `shaping.NewSliceIterator` and `Wrapper.WrapNextLine`), which is my use case.

Supporting tabs when wrapping a whole paragraph seems harder, since we have to align tabs again at the start of every line. The non exported `Output.applyTabs` should probably be used every time we "consume" a new shaped run.

### DIFF
--- a/shaping/output.go
+++ b/shaping/output.go
@@ -198,7 +198,7 @@ func (out *Output) FromFontUnit(v float32) fixed.Int26_6 {
 // and can be used to speed up line wrapping logic.
 func (out *Output) RecomputeAdvance() {
 	advance := fixed.Int26_6(0)
-	for _, g := range o.Glyphs {
+	for _, g := range out.Glyphs {
 		advance += g.Advance
 	}
 	out.Advance = advance
@@ -227,11 +227,11 @@ func (out *Output) advanceSpaceAware(paragraphDir di.Direction) fixed.Int26_6 {
 	}
 	if out.Direction.IsVertical() {
 		if lastG.Height == 0 {
-			return o.Advance - lastG.Advance
+			return out.Advance - lastG.Advance
 		}
 	} else { // horizontal
 		if lastG.Width == 0 {
-			return o.Advance - lastG.Advance
+			return out.Advance - lastG.Advance
 		}
 	}
 	return out.Advance - lastG.endLetterSpacing
@@ -248,9 +248,9 @@ func (out *Output) RecalculateAll() {
 		descent fixed.Int26_6
 	)
 
-	if o.Direction.IsVertical() {
-		for i := range o.Glyphs {
-			g := &o.Glyphs[i]
+	if out.Direction.IsVertical() {
+		for i := range out.Glyphs {
+			g := &out.Glyphs[i]
 			advance += g.Advance
 			depth := g.XOffset + g.XBearing // start of the glyph
 			if depth < descent {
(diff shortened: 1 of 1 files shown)


andydotxyz:
> Supporting tabs when wrapping a whole paragraph seems harder, since we have to align tabs again at the start of every line. 

An interesting point. Do we? In the work I have done elsewhere the tab alignment is referring to the first line (where the actual tab character was) and the others have a 0 offset.

However it is a little dependent on use-case and may not always follow. However a more complete solution is how many taps the first and "all others" indent as two options I think?

There is the complication also of where the run does not start at the beginning of the paragraph so the initial indent is less due to previous content rather than because of the tab characters...

whereswaldon (commented):
I'm not sure that I understand the expected use of the API. We define `AlignTabs` on `Line`, which is the output of line-wrapping. However, anything that modifies the advance of glyphs/runs needs to happen prior to line wrapping in order to ensure correct wrapped results. Could you provide an example of how this is meant to work in tandem with line wrapping?

benoitkugler:
> > > I'm not sure that I understand the expected use of the API. We define `AlignTabs` on `Line`, which is the output of line-wrapping. However, anything that modifies the advance of glyphs/runs needs to happen prior to line wrapping in order to ensure correct wrapped results. Could you provide an example of how this is meant to work in tandem with line wrapping?
> > 
> > 
> > I understand your concern : the `AlignTabs` method should be called _before_ line wrapping, actually on a slice of runs. I've used `Line` just as such a slice, not as the output of line wrapping, but it may be confusing. Perhaps it should be a function `AlignTabs(line []Output)` instead ?
> 
> But aren't the results of tab alignment wrapping-dependent? If there's a tab in the middle of line 2, the advance of that tab is dependent on the advance of the prior text on line two, as it needs to push the next glyph to be tabstop-aligned. I don't think we can a-priori determine how much text will be before a tab in a given line, so we can't compute the advances before wrapping. Am I missing something?
> 
> Based on my current understanding, tab expansion needs to happen during wrapping as part of processing a run to see if it can fit on the current line. I would be happy to be corrected though, as I'm not eager to mess with that part of the logic.

You are completely right, that is why I have not tackled the general case on this PR. 
But in my use case, I only wrap one line at a time. In this case, line wrapping does not break tabs (at least in LTR direction, which is already nice enough for me).

benoitkugler:
> > You are completely right, that is why I have not tackled the general case on this PR. But in my use case, I only wrap one line at a time. In this case, line wrapping does not break tabs (at least in LTR direction, which is already nice enough for me).
> 
> I'm sorry, I clearly didn't read the PR description thoroughly today. You already covered this. 🤦‍♂️
> 
> However, since expanding tabs changes the width of the line, I fail to see how it can be safe even for the line-by-line use-case. If you only call `WrapNextLine()`, you will still sometimes get a line that is packed full, and within which a tab expansion should change the wrapping decision. I believe that is possible even if the text is all the same direction. Is there some mitigation that I'm missing to handle that case?

Hum. If you apply tabs on the whole slice of runs, that is you pretend you have one "infinite" line, and then you just call to WrapNextLine to cut the line, I'm failing to see what could be wrong. Perhaps I'm simplifying the situation too much ? Have you a counter example in mind ?

PR #209, opened 2025-12-12
[shaping] support non integer maxWidth for line wrapping

### LINKED ISSUE #208: Suboptimal wrapping due to rounding
It is a bit unfortunate that `LineWrapper.WrapNextLine` and `LineWrapper.WrapParagraph` take an `int` for the maximum width. 

In one of my use case, the precision loss matters as an unexpected line break is produced by go-text. More precisely, I have a single run of text of advance 45.5 and a max width of 45.8, so that the run is broken by go-text while it should simply fit on the first line.

I think we should change the type used internally, expose two new methods taking a `fixed.26_6`, and (perhaps) deprecate the current method.

benoitkugler:
Accepts a `fixed.Int26_6` as `maxWidth`, closes #208 

~We should perhaps deprecate the `int` versions of `WrapParagraph` and `WrapNextLine`, not sure how to proceed though.~

### DIFF
--- a/shaping/wrapping.go
+++ b/shaping/wrapping.go
@@ -787,7 +787,7 @@ func (l *LineWrapper) WrapParagraph(config WrapConfig, maxWidth int, paragraph [
 	return l.WrapParagraphF(config, fixed.I(maxWidth), paragraph, runs)
 }
 
-// WrapParagraphFF is the same as [WrapParagraphF], but accepts a non integer [maxWidth].
+// WrapParagraphF is the same as [WrapParagraph], but accepts a non integer [maxWidth].
 func (l *LineWrapper) WrapParagraphF(config WrapConfig, maxWidth fixed.Int26_6, paragraph []rune, runs RunIterator) (_ []Line, truncated int) {
 	l.scratch.reset()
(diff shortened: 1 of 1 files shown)


benoitkugler:
@whereswaldon Hi Chris, any chances you could have a look at this PR ? Thank you !

shaping/wrapping.go
```diff
@@ -781,7 +781,14 @@ func (l *LineWrapper) Prepare(config WrapConfig, paragraph []rune, runs RunItera
 // See [(*LineWrapper).WrapNextLine] for a description of how [WrapConfig]'s truncation
 // features impact the wrapped text output. This method returns the quantity of runes
 // truncated by line wrapping in the [truncated] return value.
+//
+// See also [WrapParagraphF] which supports a decimal [maxWidth].
 func (l *LineWrapper) WrapParagraph(config WrapConfig, maxWidth int, paragraph []rune, runs RunIterator) (_ []Line, truncated int) {
+	return l.WrapParagraphF(config, fixed.I(maxWidth), paragraph, runs)
+}
+
+// WrapParagraphFF is the same as [WrapParagraphF], but accepts a non integer [maxWidth].
```
whereswaldon: The name in this comment doesn't match the, implementation.

Issue #216, opened 2026-01-16
Font scanning fails in Flatpak due to broken symlinks. (Broken Chinese characters in flatpak)


alexballas:
When running applications using go-text/typesetting inside Flatpak, `fontscan.UseSystemFonts()` fails with an error like: 
```
stat /run/host/fonts/cMap/Identity-UTF16-H: no such file or directory
```

Chinese and other non-latin fonts look broken as a result of that.

This happens because some font directories contain broken symlinks (e.g., cMap/Identity-UTF16-H pointing to ../../poppler/cMap/... which may be inaccessible inside Flatpak). When `os.Stat()` encounters these broken symlinks, it returns an error that causes the entire font scan to fail.

In `fontscan/scandir.go`, the `scanDirectory()` function returns the error from `os.Stat()` on line 30, which aborts the entire font scan.

PR #217, opened 2026-01-16
Modify error return to nil in scandir.go (Fixes broken Chinese characters in flatpak)

alexballas:
Change error handling to return nil instead of an error to avoid causing the entire font scan to fail.

### DIFF
--- a/fontscan/scandir.go
+++ b/fontscan/scandir.go
@@ -1,6 +1,8 @@
 package fontscan
 
 import (
+	"errors"
+	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
@@ -27,7 +29,11 @@ func (dst *footprintScanner) scanDirectory(logger Logger, dir string, visited ma
 		// load the information, following potential symoblic links
 		info, err := os.Stat(path)
 		if err != nil {
-			return err
+			if errors.Is(err, os.ErrNotExist) {
+				logger.Printf("skipping dead link or missing file: %q", path)
+				return nil
+			}
+			return fmt.Errorf("failed to stat %q: %w", path, err)
 		}
 
 		// always ignore files which should never be font files


benoitkugler:
Thank you for this PR ! 
I am wondering if we should log a warning, something like 
`logger.Printf("error walking font directory %q: %v", path, err)`

dweymouth:
> Thank you for this PR ! I am wondering if we should log a warning, something like `logger.Printf("error walking font directory %q: %v", path, err)`

Or maybe just a code comment explaining why we're ignoring the error? We expect this "error" when running in the Flatpak sandbox, so it might pollute logs if we always log it?

alexballas:
flatpaks are mainly designed for GUI apps, so no flatpak user will ever see this error.

This error would be for the non-flatpak users. However even in this case and depending how many broken links you have, it would spam quite a bit.

alexballas:
@andydotxyz this is indeed broken links as highlighted in my comment here https://github.com/go-text/typesetting/issues/216#issue-3822666675 

During my tests I added custom logging to confirm.

```
alex@zeus:~$ flatpak run --command=bash app.go2tv.go2tv
[📦 app.go2tv.go2tv ~]$ ls -al /run/host/fonts/cMap/
total 8
drwxr-xr-x 2 nfsnobody nfsnobody 4096 Aug 27  2024 .
drwxr-xr-x 8 nfsnobody nfsnobody 4096 Aug 27  2024 ..
lrwxrwxrwx 1 nfsnobody nfsnobody   35 Feb  9  2023 Identity-UTF16-H -> ../../poppler/cMap/Identity-UTF16-H
[📦 app.go2tv.go2tv ~]$ cat /run/host/fonts/cMap/Identity-UTF16-H
cat: /run/host/fonts/cMap/Identity-UTF16-H: No such file or directory
[📦 app.go2tv.go2tv ~]$ 
```

andydotxyz:
My apologies I did not mean to doubt that.
What I meant is that ignoring all the possible errors returned by that Stat call might not be the right solution - instead the directory traversal should perhaps filter out irrelevant errors itself rather than having them never reported.

alexballas:
cool, I pushed an updated version.

Issue #210, opened 2025-12-12
Glyph struct optimisation


benoitkugler:
Hi,

I was wondering how you would consider a breaking change cleaning and optimizing the `shaping.Glyph` struct.
There are two changes I would like to make :

-  replacing the fields XAdvance and YAdvance by an unique Advance  field. XAdvance and YAdvance are never used together : XAdvance is for horizontal text, YAdvance for vertical text. With this change, we could also simplify our internal logic and remove switches on the direction.

-  as once suggested by Chris, we could reduce the size of the RuneCount and GlyphCount fields to use uint8. In practice, theses numbers never exceed (say) 10, so using a full int is a bit of a waste.

whereswaldon:
> I'm pretty sure harfbuzz only sets the advance of the shaping direction, because the advance (in general) is influenced by surrounding glyphs and opentype layout. As opposed to glyph Width and Height. In my understanding, the advance of a glyph may vary between shaping anyway and should not be cached.

They probably are contextual, but it's worth noting that Harfbuzz offers both in the output: 

https://github.com/go-text/typesetting/blob/9645b81b4737826488ff2b5bce962bf742d0b931/shaping/shaping.go#L135

Maybe the expectation is that you'll only access one of those fields though.

benoitkugler:
> > I'm pretty sure harfbuzz only sets the advance of the shaping direction, because the advance (in general) is influenced by surrounding glyphs and opentype layout. As opposed to glyph Width and Height. In my understanding, the advance of a glyph may vary between shaping anyway and should not be cached.
> 
> They probably are contextual, but it's worth noting that Harfbuzz offers both in the output:
> 
> [typesetting/shaping/shaping.go](https://github.com/go-text/typesetting/blob/9645b81b4737826488ff2b5bce962bf742d0b931/shaping/shaping.go#L135)
> 
> Line 135 in [9645b81](/go-text/typesetting/commit/9645b81b4737826488ff2b5bce962bf742d0b931)
> 
>  glyphs[i].YAdvance = fixed.I(int(t.buf.Pos[i].YAdvance)) >> scaleShift 
> Maybe the expectation is that you'll only access one of those fields though.

You're right. It not obvious at a first glance of the harfbuzz package that only one of the field is used at a time. I'll need to double check that.
(Having said that, I fail to understand how YAdvance could be useful in horizontal layout, since there already is the Yoffset field to perform cross alignement.)

benoitkugler:
[...]

PR #219, opened 2026-01-20
[shaping] add Advance; deprecate some other fields

benoitkugler:
This implements the transition path discussed in #210.

I've tried and make the minimal changes so that tests pass. As proposed, this PR 

- adds methods `Glyph.TextIndex() int`, `Glyph.RunesCount() int`, `Glyph.GlyphsCount() int` 
- adds a field `Glyph.Advance`
- deprecates `XAdvance`, `YAdvance`, `ClusterIndex`, `RuneCount`, `GlyphCount` fields

### DIFF
--- a/shaping/output.go
+++ b/shaping/output.go
@@ -26,11 +26,19 @@ type Glyph struct {
 	// YBearing is the distance between the dot (with offset applied) and
 	// the top of the glyph content, typically positive
 	YBearing fixed.Int26_6
+	// Advance is the distance between the current dot (without offset applied) and the next dot.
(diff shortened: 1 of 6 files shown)


benoitkugler:
@andydotxyz Any chance you could have a look at these changes? It would be nice to publish them and start updating users codebases.

andydotxyz (approved):
Sorry this escaped me in a holiday catchup.

Thanks for putting the effort into a migration path.

Perhaps we could/should now tag a release and note that it's the last one supporting these deprecated fields?

Issue #218, opened 2026-01-16
Need to get a `GlyphSVG` instead of `GlyphColor`


rcoreilly:
The addition of support for the COLR table in #196 breaks cogentcore's emoji font rendering, which is setup to process the `GlyphSVG` data. In `Face.GlyphData`, the COLR table has precedence, and there is no way to influence this behavior. 

One strategy would be to add a field in `GlyphColor` that points to the SVG data if it is available? This is what GlyphBitmap does with the Outline field.

PR #220, opened 2026-01-20
[font] expose more granular glyph data methods

benoitkugler:
This is a fix for #218 : it exposes a specific method for each kind of glyphs we support.

(cc @rcoreilly)


whereswaldon (approved):
Looks great to me; thank you!

Issue #190, opened 2025-04-20
Question: why is the default `HarfBuzzShaper` cache size 0


hajimehoshi:
Yesterday, I found that `HarfBuzzShaper`'s cache size was 0 and `NewFont` was called every time different texts are rendered: 
* https://github.com/hajimehoshi/guigui/issues/64
* https://github.com/hajimehoshi/ebiten/issues/3228

This issue could be prevented by settings some cache size. I was wondering why 0 is the initial value. Would it be possible to have initial cache size to avoid future troubles like me?

hajimehoshi:
> I guess the reason is that we do not have a constructor for HarfBuzzShaper and the zero value for the fontLRU cache has 0 as its maxSize.

If this struct needs to have a non-zero cache by default, I'd have an integer member substracted by the default value so that the initial value (0) indicates the default value.

benoitkugler:
I actually thought about it, but what would `SetCacheSize` do then ? It would not do what the user expects

hajimehoshi:
> what would SetCacheSize do then ? 

This function would update the member variable with the argument minus the default size.

PR #221, opened 2026-01-20
[shaping] adjust default cache size

benoitkugler:
This is a fix for #190 : the `HarfbuzzShaper` zero value now has a non empty font cache.
 
@whereswaldon @hajimehoshi  Maybe you have an educated guess for the value of `defaultFontCacheSize` ?


whereswaldon (approved):
I don't have a good intuition about this default sizing, but 32 seems like a fine starting point. Most applications I've worked with didn't actively use more than 32 fonts.

PR #225, opened 2026-02-03
 Update Harfbuzz implementation; upgrade to Unicode 17

benoitkugler:
This PR ports the latest changes in the Harfbuzz C++ implementation (roughly v12.3.0).
It also bumps our Unicode version to 17, and upgrades our Unicode Line Breaking implementation (in the segmenter package).

There is no API change, expect for a change in the method `Face.GlyphVOrigin` whose signature is now `GlyphVOrigin(glyph GID) (x, y float32)`. This method is used internally in Harfbuzz, and this change should not affect end users.


andydotxyz (commented):
> There is no API change

I could be wrong but it looks like there are some public API changes - in the tables packages it looks like a number of exported functions were removed (GetDelta/Evaluate).

The API change you mention:
```diff
- func (f *Face) GlyphVOrigin(glyph GID) (x, y int32, found bool) {
+ func (f *Face) GlyphVOrigin(glyph GID) (x, y float32) {
```

is a public API and not internal only. Should much of this have been in an `internal` package, or are they APIs useful publicly as well?

benoitkugler:
> Ultimately, I cannot intelligently review all of this. I have tried to review the non-generated code changes, and I see nothing that concerns me. However, the magnitude of these changes makes careful review impossible. I would really prefer to avoid PRs like this in the future.

I understand your point. Mixing Unicode changes with other Harfbuzz stuff was a bad idea. I'm sorry for that.

If it is important to you, I can still try and split this PR in several smaller ones.

PR #235, opened 2026-02-25
Glyph struct pruning

### LINKED ISSUE #210: Glyph struct optimisation

benoitkugler:
This is the breaking change discussed in #210, enabling memory savings.

Closes #210

### DIFF
--- a/shaping/output.go
+++ b/shaping/output.go
@@ -36,24 +36,18 @@ type Glyph struct {
 	// below the horizontal line at the dot
 	XOffset, YOffset fixed.Int26_6
 
-	// ClusterIndex is the lowest rune index of all runes shaped into
-	// this glyph cluster. All glyphs sharing the same cluster value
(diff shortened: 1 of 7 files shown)


benoitkugler:
I've added a v0.3.4 tag including the latest changes.

andydotxyz:
Thanks, I updated renderer and tagged v0.2.1 accordingly

PR #243, opened 2026-03-18
[font] expose the flavor

benoitkugler:
We already compute this flag in the lower-level `loader` package: this PR surfaces this information.


andydotxyz (approved):
Cool

PR #245, opened 2026-03-27
[shaping] better face selection for mixed RTL/LTR with marks

benoitkugler:
When shaping the string `\u200fabc`, out segmentation process produces two runs (`Input`) with correct direction and script, but with _different_ fonts. This is suboptimal : we should not change font for runes such as  \u200f (a RTL mark).

This issue is that our font selection function only works one `Input` at a time, whereas a "cross" run analysis is required here.

This PR aims at fixing this issue, by back propagating the selected font to previous `Input`s.

This is a minor change which should not impact the majority of use cases.


whereswaldon (approved):
Thank you! This is a tricky case indeed.

PR #246, opened 2026-03-27
[font] Support for EBDT format 1

### LINKED ISSUE #244: EBDT image format 1 is not supported
We found that go-text/typesetting didn't handle EBDT format correctly. The discussion is at https://github.com/hajimehoshi/ebiten/discussions/3425#discussioncomment-16336073.

The below report was made by Claude Code:

## Summary

`BitmapSizes()` returns an empty slice for fonts that use EBDT image format 1 (small metrics, byte-aligned data), because `parseBitmapDataMetrics` in `font/bitmaps.go` explicitly skips it as "not implemented". This affects common CJK fonts like SimSun (simsun.ttc).

## Reproduction

```go
f, _ := os.Open("simsun.ttc")
loaders, _ := opentype.NewLoaders(f)
font, _ := font.NewFont(loaders[0])
fmt.Println(font.BitmapSizes()) // [] (empty)
```

SimSun's EBLC/EBDT tables contain 6 bitmap strikes (ppem 12–17) with 66 index subtables covering ~28,000 glyphs. All subtables use image format 1.

## Cause

In `font/bitmaps.go:431-433`:

```go
switch imageFormat {
case 1, 6, 7, 8, 9:
    return bitmapImage{}, fmt.Errorf("valid but currently not implemented bitmap image format: %d", imageFormat)
```

This error propagates through `newBitmapSubtable` → `newBitmap` → `loadBitmap` → `selectBitmapTable`, causing the entire bitmap table to be discarded.

## Proposed fix

Format 1 (small metrics, byte-aligned data) has the same structure as format 2 (small metrics, bit-aligned data): `SmallGlyphMetrics` followed by image data. The only difference is row alignment. The existing `BitmapData2` struct and `ParseBitmapData2` should work for parsing format 1 as well.

```go
case 1, 2:
    data, _, err := tables.ParseBitmapData2(imageData)
    return bitmapImage{metrics: data.SmallGlyphMetrics, image: data.Image}, err
```

The byte-aligned vs bit-aligned distinction matters at rendering time, not at parsing time. If the renderer needs to distinguish them, the image format could be stored in `bitmapImage`.

benoitkugler:
This PR adds support for bitmap black and white glyphs with Format1.

Since the rendering logic should be slightly different than other black and white format already supported, I have added a new format `BlackAndWhiteByteAligned`.

Closes #244

### DIFF
--- a/font/opentype/tables/glyphs_bitmap_src.go
+++ b/font/opentype/tables/glyphs_bitmap_src.go
@@ -163,7 +163,7 @@ type BigGlyphMetrics struct {
 
 // Format 1: small metrics, byte-aligned data
 // Format 2: small metrics, bit-aligned data
-type BitmapData1or2 struct {
+type BitmapData1Or2 struct {
 	SmallGlyphMetrics
 	Image []byte `arrayCount:"ToEnd"`
 }
--- a/font/opentype/tables/glyphs_bitmap_src.go
+++ b/font/opentype/tables/glyphs_bitmap_src.go
@@ -161,8 +161,9 @@ type BigGlyphMetrics struct {
 	vertAdvance  uint8 // Vertical advance width in pixels.
 }
 
+// Format 1: small metrics, byte-aligned data
 // Format 2: small metrics, bit-aligned data
-type BitmapData2 struct {
+type BitmapData1or2 struct {
 	SmallGlyphMetrics
 	Image []byte `arrayCount:"ToEnd"`
 }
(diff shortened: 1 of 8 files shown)


font/opentype/tables/glyphs_bitmap_src.go
```diff
@@ -161,8 +161,9 @@ type BigGlyphMetrics struct {
 	vertAdvance  uint8 // Vertical advance width in pixels.
 }
 
+// Format 1: small metrics, byte-aligned data
 // Format 2: small metrics, bit-aligned data
-type BitmapData2 struct {
+type BitmapData1or2 struct {
```
hajimehoshi: `BitmapData1Or2` sounds slighly better (to me)

whereswaldon (approved):
Thanks for addressing this so quickly!

font/opentype/tables/glyphs_bitmap_src.go
benoitkugler: Done, thank you.

benoitkugler:
Waiting for @andydotxyz approval since there is an API change.

andydotxyz (approved):

hajimehoshi:
Thanks!

PR #248, opened 2026-04-10
Don't ignore face change for whitespace

### LINKED ISSUE #6216: Android: Unicode character "✘ " (heavy ballot x + space) displayed as "✘�"
### Checklist

- [x] I have searched the [issue tracker](https://github.com/fyne-io/fyne/issues) for open issues that relate to the same problem, before opening a new one.
- [x] This issue only relates to a single bug. I will open new issues for any other problems.

### Describe the bug

The character "✘" (heavy ballot x) is displayed as "✘�" on Android. It's strange that "✔" (heavy check mark) is displayed correctly. Both characters use three bytes in UTF-8 (`0xe2 9c 98` and `0xe2 9c 94`).

### How to reproduce

Build, install, and run example app on Android.

### Screenshots

Image

### Example code

```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

func main() {
	a := app.New()
	w := a.NewWindow("Unicode test")

	label := widget.NewLabel(`✘ heavy ballot x (U+2718)
✔ heavy check mark (U+2714)`)
	w.SetContent(label)
	w.ShowAndRun()
}
```

### Fyne version

develop branch

### Go compiler version

1.26.1

### Operating system and version

Android 13

### Additional Information

_No response_

MaxGyver83:
The assumption "that all fonts have the ASCII space" doesn't hold for Android. P.e. the font `/system/fonts/NotoSansSymbols-Regular-Subsetted.ttf` on my Android device (OnePlus Nord 2 5G), which is used to display "✘", doesn't contain a space character/glyph.

This fixes <https://github.com/fyne-io/fyne/issues/6216>.

### DIFF
--- a/shaping/input.go
+++ b/shaping/input.go
@@ -434,11 +434,22 @@ func splitByFace(input Input, availableFaces Fontmap, buffer []Input, isLast boo
 				// add the rune to the current input
 				continue
 			}
-			if currentInput.Face != nil && isSpace {
-				// We must check if the current face contains an ASCII space.
-				// P.e. NotoSansSymbols-Regular-Subsetted.ttf (on Android) doesn't.
-				if _, ok := currentInput.Face.NominalGlyph(rune(' ')); ok {
-					continue
+			if isSpace {
+				if currentInput.Face == nil {
+					for j := i + 1; j < input.RunEnd; j++ {
+						ignore, isSpace := ignoreFaceChange(input.Text[j])
+						if !ignore && !isSpace {
+							currentInput.Face = availableFaces.ResolveFace(input.Text[j])
+							break
+						}
+					}
+				}
+				if currentInput.Face != nil {
+					// We must check if the current face contains an ASCII space.
+					// P.e. NotoSansSymbols-Regular-Subsetted.ttf (on Android) doesn't.
+					if _, ok := currentInput.Face.NominalGlyph(rune(' ')); ok {
+						continue
+					}
 				}
 			}
 		}
--- a/shaping/input.go
+++ b/shaping/input.go
@@ -428,9 +428,16 @@ func splitByFace(input Input, availableFaces Fontmap, buffer []Input, isLast boo
 		// We can safely ignore characters if we have a face or if there is more text,
 		// but we must force the choice of a face if we still don't have one and we reach
 		// the final rune. Otherwise strings like all-whitespace are never assigned a face.
-		if ignoreFaceChange(r) && (currentInput.Face != nil || !isLast || i < input.RunEnd-1) {
-			// add the rune to the current input
-			continue
+		if currentInput.Face != nil || !isLast || i < input.RunEnd-1 {
+			if ignoreFaceChange(r) {
+				// add the rune to the current input
+				continue
+			}
+			if currentInput.Face != nil && isSpace(r) {
+				if _, ok := currentInput.Face.NominalGlyph(rune(' ')); ok {
+					continue
+				}
+			}
 		}
 
 		// select the first font supporting r
@@ -471,12 +478,6 @@ func splitByFace(input Input, availableFaces Fontmap, buffer []Input, isLast boo
 // ignoreFaceChange returns `true` is the given rune should not trigger
 // a change of font.
 //
-// We don't want space characters to affect font selection; in general,
-// it's always wrong to select a font just to render a space.
-// We assume that all fonts have the ASCII space, and for other space
-// characters if they don't, HarfBuzz will compatibility-decompose them
-// to ASCII space...
-//
 // We don't want to change fonts for line or paragraph separators.
 //
 // Finaly, we also don't change fonts for what Harfbuzz consider
@@ -494,6 +495,18 @@ func ignoreFaceChange(r rune) bool {
 		g == ucd.Zl || // line separator
 		g == ucd.Zp || // paragraph separator
 		harfbuzz.IsDefaultIgnorable(r)
+}
+
+// isSpace returns `true` if the given rune is a space character.
+//
+// We don't want space characters to affect font selection; in general,
+// it's always wrong to select a font just to render a space.
+// We can check if a font has the ASCII space, and for other space
+// characters if they don't, HarfBuzz will compatibility-decompose them
+// to ASCII space...
+func isSpace(r rune) bool {
+	g := ucd.LookupGeneralCategory(r)
+	return g == ucd.Zs && r != '\u1680'
 }
 
 // enforceLang makes sure the returned language is compatible with
--- a/shaping/input.go
+++ b/shaping/input.go
@@ -505,7 +505,7 @@ func ignoreFaceChange(r rune) (ignore, isSpace bool) {
 			g == ucd.Zl || // line separator
 			g == ucd.Zp || // paragraph separator
 			harfbuzz.IsDefaultIgnorable(r),
-		g == ucd.Zs && r != '\u1680'
+		g == ucd.Zs && r != '\u1680' // space separator != OGHAM SPACE MARK
 }
 
 // enforceLang makes sure the returned language is compatible with
--- a/shaping/input.go
+++ b/shaping/input.go
@@ -493,7 +493,6 @@ func ignoreFaceChange(r rune) bool {
 		g == ucd.Cs || // surrogate
 		g == ucd.Zl || // line separator
 		g == ucd.Zp || // paragraph separator
-		(g == ucd.Zs && r != '\u1680') || // space separator != OGHAM SPACE MARK
 		harfbuzz.IsDefaultIgnorable(r)
 }
 
--- a/shaping/input.go
+++ b/shaping/input.go
@@ -429,11 +429,14 @@ func splitByFace(input Input, availableFaces Fontmap, buffer []Input, isLast boo
 		// but we must force the choice of a face if we still don't have one and we reach
 		// the final rune. Otherwise strings like all-whitespace are never assigned a face.
 		if currentInput.Face != nil || !isLast || i < input.RunEnd-1 {
-			if ignoreFaceChange(r) {
+			ignore, isSpace := ignoreFaceChange(r)
+			if ignore {
 				// add the rune to the current input
 				continue
 			}
-			if currentInput.Face != nil && isSpace(r) {
+			if currentInput.Face != nil && isSpace {
+				// We must check if the current face contains an ASCII space.
+				// P.e. NotoSansSymbols-Regular-Subsetted.ttf (on Android) doesn't.
 				if _, ok := currentInput.Face.NominalGlyph(rune(' ')); ok {
 					continue
 				}
@@ -476,7 +479,14 @@ func splitByFace(input Input, availableFaces Fontmap, buffer []Input, isLast boo
 }
 
 // ignoreFaceChange returns `true` is the given rune should not trigger
-// a change of font.
+// a change of font and a second bool that indicates if the rune is a space.
+//
+// We don't want space characters to affect font selection; in general,
+// it's always wrong to select a font just to render a space.
+// We assume that most fonts have the ASCII space, and for other space
+// characters if they don't, HarfBuzz will compatibility-decompose them
+// to ASCII space... We return isSpace because we want to be able to
+// force a space to be rendered with a different font if necessary.
 //
 // We don't want to change fonts for line or paragraph separators.
 //
@@ -488,25 +498,14 @@ func splitByFace(input Input, availableFaces Fontmap, buffer []Input, isLast boo
 // https://bugzilla.gnome.org/show_bug.cgi?id=701652
 // https://bugzilla.gnome.org/show_bug.cgi?id=781123
 // for more details.
-func ignoreFaceChange(r rune) bool {
+func ignoreFaceChange(r rune) (ignore, isSpace bool) {
 	g := ucd.LookupGeneralCategory(r)
 	return g == ucd.Cc || // control
-		g == ucd.Cs || // surrogate
-		g == ucd.Zl || // line separator
-		g == ucd.Zp || // paragraph separator
-		harfbuzz.IsDefaultIgnorable(r)
-}
-
-// isSpace returns `true` if the given rune is a space character.
-//
-// We don't want space characters to affect font selection; in general,
-// it's always wrong to select a font just to render a space.
-// We can check if a font has the ASCII space, and for other space
-// characters if they don't, HarfBuzz will compatibility-decompose them
-// to ASCII space...
-func isSpace(r rune) bool {
-	g := ucd.LookupGeneralCategory(r)
-	return g == ucd.Zs && r != '\u1680'
+			g == ucd.Cs || // surrogate
+			g == ucd.Zl || // line separator
+			g == ucd.Zp || // paragraph separator
+			harfbuzz.IsDefaultIgnorable(r),
+		g == ucd.Zs && r != '\u1680'
 }
 
 // enforceLang makes sure the returned language is compatible with
--- a/shaping/input_test.go
+++ b/shaping/input_test.go
@@ -17,7 +17,7 @@ func Test_ignoreFaceChange(t *testing.T) {
 		args rune
 		want bool
 	}{
-		{' ', true},
+		{' ', false},
 		{'a', false},
 		{'\n', true},
 		{'\r', true},
@@ -139,8 +139,13 @@ func TestSplitByFontGlyphs(t *testing.T) {
 				},
 				{
 					Text:     []rune("aaa AAA "),
-					RunStart: 4, RunEnd: 8,
+					RunStart: 4, RunEnd: 7,
 					Face: upperFont,
+				},
+				{
+					Text:     []rune("aaa AAA "),
+					RunStart: 7, RunEnd: 8,
+					Face: lowerFont,
 				},
 			},
 		},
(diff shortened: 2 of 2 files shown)


andydotxyz (approved):
This looks good to me - but would like anyone more familiar to check it's the best way to solve the problem and we're not introducing some big inefficiency?

MaxGyver83:
Oh, I have forgotten to update the tests. While doing this, I noticed that with this change, the font is switched to latin for spaces in arabic text. Here it would be better to use the arabic font for spaces, too (Amiri-Regular.ttf contains `U+20`).

benoitkugler:
> Oh, I have forgotten to update the tests. While doing this, I noticed that with this change, the font is switched to latin for spaces in arabic text. Here it would be better to use the arabic font for spaces, too (Amiri-Regular.ttf contains `U+20`).

That is precisely why I think this change is not the right fix.

You seem to use a subsetted font. Could you check the original one does not contain the space glyph ?

MaxGyver83:
> > Oh, I have forgotten to update the tests. While doing this, I noticed that with this change, the font is switched to latin for spaces in arabic text. Here it would be better to use the arabic font for spaces, too (Amiri-Regular.ttf contains `U+20`).
> 
> That is precisely why I think this change is not the right fix.
> 
> You seems to use a subsetted font. Could you check the original one does not contain the space glyph ?

Probably yes, but it seems not to be available on my device.

Maybe this would help!?

```diff
diff --git a/shaping/input.go b/shaping/input.go
index 60d46e7..8f405dc 100644
--- a/shaping/input.go
+++ b/shaping/input.go
@@ -433,6 +433,12 @@ func splitByFace(input Input, availableFaces Fontmap, buffer []Input, isLast boo
                        continue
                }

+               if currentInput.Face != nil {
+                       if _, ok := currentInput.Face.NominalGlyph(r); ok {
+                               continue
+                       }
+               }
+
                // select the first font supporting r
                selectedFace := availableFaces.ResolveFace(r)
```

(This checks first if the current font can be kept for the current rune. Otherwise, it calls the (probably more expensive) `ResolveFace`.)

MaxGyver83:
> You seems to use a subsetted font. Could you check the original one does not contain the space glyph ?

These are the available fonts:

```sh
$ adb shell ls /system/fonts/
AndroidClock.ttf
CarroisGothicSC-Regular.ttf
ComingSoon.ttf
CutiveMono.ttf
DancingScript-Bold.ttf
DancingScript-Regular.ttf
DroidSans-Bold.ttf
DroidSans.ttf
DroidSansMono.ttf
FZZWXBTOT_Uni.ttf
MyanmarZg.ttf
NotoColorEmoji.ttf
NotoColorEmojiFlags.ttf
NotoColorEmojiLegacy.ttf
NotoNaskhArabic-Bold.ttf
NotoNaskhArabic-Regular.ttf
NotoNaskhArabicUI-Bold.ttf
NotoNaskhArabicUI-Regular.ttf
NotoSansAdlam-VF.ttf
NotoSansAhom-Regular.otf
NotoSansAnatolianHieroglyphs-Regular.otf
NotoSansArmenian-VF.ttf
NotoSansAvestan-Regular.ttf
NotoSansBalinese-Regular.ttf
NotoSansBamum-Regular.ttf
NotoSansBassaVah-Regular.otf
NotoSansBatak-Regular.ttf
NotoSansBengali-VF.ttf
NotoSansBengaliUI-VF.ttf
NotoSansBhaiksuki-Regular.otf
NotoSansBrahmi-Regular.ttf
NotoSansBuginese-Regular.ttf
NotoSansBuhid-Regular.ttf
NotoSansCJK-Regular.ttc
NotoSansCanadianAboriginal-Regular.ttf
NotoSansCarian-Regular.ttf
NotoSansChakma-Regular.otf
NotoSansCham-Bold.ttf
NotoSansCham-Regular.ttf
NotoSansCherokee-Regular.ttf
NotoSansCoptic-Regular.ttf
NotoSansCuneiform-Regular.ttf
NotoSansCypriot-Regular.ttf
NotoSansDeseret-Regular.ttf
NotoSansDevanagari-VF.ttf
NotoSansDevanagariUI-VF.ttf
NotoSansEgyptianHieroglyphs-Regular.ttf
NotoSansElbasan-Regular.otf
NotoSansEthiopic-VF.ttf
NotoSansGeorgian-VF.ttf
NotoSansGlagolitic-Regular.ttf
NotoSansGothic-Regular.ttf
NotoSansGrantha-Regular.ttf
NotoSansGujarati-Bold.ttf
NotoSansGujarati-Regular.ttf
NotoSansGujaratiUI-Bold.ttf
NotoSansGujaratiUI-Regular.ttf
NotoSansGunjalaGondi-Regular.otf
NotoSansGurmukhi-VF.ttf
NotoSansGurmukhiUI-VF.ttf
NotoSansHanifiRohingya-Regular.otf
NotoSansHanunoo-Regular.ttf
NotoSansHatran-Regular.otf
NotoSansHebrew-Bold.ttf
NotoSansHebrew-Regular.ttf
NotoSansImperialAramaic-Regular.ttf
NotoSansInscriptionalPahlavi-Regular.ttf
NotoSansInscriptionalParthian-Regular.ttf
NotoSansJavanese-Regular.otf
NotoSansKaithi-Regular.ttf
NotoSansKannada-VF.ttf
NotoSansKannadaUI-VF.ttf
NotoSansKayahLi-Regular.ttf
NotoSansKharoshthi-Regular.ttf
NotoSansKhmer-VF.ttf
NotoSansKhmerUI-Bold.ttf
NotoSansKhmerUI-Regular.ttf
NotoSansKhojki-Regular.otf
NotoSansLao-Bold.ttf
NotoSansLao-Regular.ttf
NotoSansLaoUI-Bold.ttf
NotoSansLaoUI-Regular.ttf
NotoSansLepcha-Regular.ttf
NotoSansLimbu-Regular.ttf
NotoSansLinearA-Regular.otf
NotoSansLinearB-Regular.ttf
NotoSansLisu-Regular.ttf
NotoSansLycian-Regular.ttf
NotoSansLydian-Regular.ttf
NotoSansMalayalam-VF.ttf
NotoSansMalayalamUI-VF.ttf
NotoSansMandaic-Regular.ttf
NotoSansManichaean-Regular.otf
NotoSansMarchen-Regular.otf
NotoSansMasaramGondi-Regular.otf
NotoSansMedefaidrin-VF.ttf
NotoSansMeeteiMayek-Regular.ttf
NotoSansMeroitic-Regular.otf
NotoSansMiao-Regular.otf
NotoSansModi-Regular.ttf
NotoSansMongolian-Regular.ttf
NotoSansMro-Regular.otf
NotoSansMultani-Regular.otf
NotoSansMyanmar-Bold.otf
NotoSansMyanmar-Medium.otf
NotoSansMyanmar-Regular.otf
NotoSansMyanmarUI-Bold.otf
NotoSansMyanmarUI-Medium.otf
NotoSansMyanmarUI-Regular.otf
NotoSansNKo-Regular.ttf
NotoSansNabataean-Regular.otf
NotoSansNewTaiLue-Regular.ttf
NotoSansNewa-Regular.otf
NotoSansOgham-Regular.ttf
NotoSansOlChiki-Regular.ttf
NotoSansOldItalic-Regular.ttf
NotoSansOldNorthArabian-Regular.otf
NotoSansOldPermic-Regular.otf
NotoSansOldPersian-Regular.ttf
NotoSansOldSouthArabian-Regular.ttf
NotoSansOldTurkic-Regular.ttf
NotoSansOriya-Bold.ttf
NotoSansOriya-Regular.ttf
NotoSansOriyaUI-Bold.ttf
NotoSansOriyaUI-Regular.ttf
NotoSansOsage-Regular.ttf
NotoSansOsmanya-Regular.ttf
NotoSansPahawhHmong-Regular.otf
NotoSansPalmyrene-Regular.otf
NotoSansPauCinHau-Regular.otf
NotoSansPhagsPa-Regular.ttf
NotoSansPhoenician-Regular.ttf
NotoSansRejang-Regular.ttf
NotoSansRunic-Regular.ttf
NotoSansSamaritan-Regular.ttf
NotoSansSaurashtra-Regular.ttf
NotoSansSharada-Regular.otf
NotoSansShavian-Regular.ttf
NotoSansSinhala-VF.ttf
NotoSansSinhalaUI-VF.ttf
NotoSansSoraSompeng-Regular.otf
NotoSansSoyombo-VF.ttf
NotoSansSundanese-Regular.ttf
NotoSansSylotiNagri-Regular.ttf
NotoSansSymbols-Regular-Subsetted.ttf
NotoSansSymbols-Regular-Subsetted2.ttf
NotoSansSyriacEastern-Regular.ttf
NotoSansSyriacEstrangela-Regular.ttf
NotoSansSyriacWestern-Regular.ttf
NotoSansTagalog-Regular.ttf
NotoSansTagbanwa-Regular.ttf
NotoSansTaiLe-Regular.ttf
NotoSansTaiTham-Regular.ttf
NotoSansTaiViet-Regular.ttf
NotoSansTakri-VF.ttf
NotoSansTamil-VF.ttf
NotoSansTamilUI-VF.ttf
NotoSansTelugu-VF.ttf
NotoSansTeluguUI-VF.ttf
NotoSansThaana-Bold.ttf
NotoSansThaana-Regular.ttf
NotoSansThai-Bold.ttf
NotoSansThai-Regular.ttf
NotoSansThaiUI-Bold.ttf
NotoSansThaiUI-Regular.ttf
NotoSansTifinagh-Regular.otf
NotoSansUgaritic-Regular.ttf
NotoSansVai-Regular.ttf
NotoSansWancho-Regular.otf
NotoSansWarangCiti-Regular.otf
NotoSansYi-Regular.ttf
NotoSerif-Bold.ttf
NotoSerif-BoldItalic.ttf
NotoSerif-Italic.ttf
NotoSerif-Regular.ttf
NotoSerifArmenian-VF.ttf
NotoSerifBengali-VF.ttf
NotoSerifCJK-Regular.ttc
NotoSerifDevanagari-VF.ttf
NotoSerifDogra-Regular.ttf
NotoSerifEthiopic-VF.ttf
NotoSerifGeorgian-VF.ttf
NotoSerifGujarati-VF.ttf
NotoSerifGurmukhi-VF.ttf
NotoSerifHebrew-Bold.ttf
NotoSerifHebrew-Regular.ttf
NotoSerifKannada-VF.ttf
NotoSerifKhmer-Bold.otf
NotoSerifKhmer-Regular.otf
NotoSerifLao-Bold.ttf
NotoSerifLao-Regular.ttf
NotoSerifMalayalam-VF.ttf
NotoSerifMyanmar-Bold.otf
NotoSerifMyanmar-Regular.otf
NotoSerifNyiakengPuachueHmong-VF.ttf
NotoSerifSinhala-VF.ttf
NotoSerifTamil-VF.ttf
NotoSerifTelugu-VF.ttf
NotoSerifThai-Bold.ttf
NotoSerifThai-Regular.ttf
NotoSerifTibetan-VF.ttf
NotoSerifYezidi-VF.ttf
OPSans-En-Regular.ttf
OSans-Ext-Regular.otf
OplusOSUI-XThin.ttf
Roboto-Regular.ttf
RobotoStatic-Regular.ttf
SourceSansPro-Bold.ttf
SourceSansPro-BoldItalic.ttf
SourceSansPro-Italic.ttf
SourceSansPro-Regular.ttf
SourceSansPro-SemiBold.ttf
SourceSansPro-SemiBoldItalic.ttf
SysFont-Myanmar.ttf
SysFont-Regular.ttf
SysFont-Static-Regular.ttf
SysSans-En-Regular.ttf
SysSans-Hans-Regular.ttf
SysSans-Hant-Regular.ttf
```

No `NotoSansSymbols-Regular.ttf`.

MaxGyver83:
> Maybe this would help!?
> 
> ```diff
> diff --git a/shaping/input.go b/shaping/input.go
> index 60d46e7..8f405dc 100644
> --- a/shaping/input.go
> +++ b/shaping/input.go
> @@ -433,6 +433,12 @@ func splitByFace(input Input, availableFaces Fontmap, buffer []Input, isLast boo
>                         continue
>                 }
> 
> +               if currentInput.Face != nil {
> +                       if _, ok := currentInput.Face.NominalGlyph(r); ok {
> +                               continue
> +                       }
> +               }
> +
>                 // select the first font supporting r
>                 selectedFace := availableFaces.ResolveFace(r)
> ```
> 
> (This checks first if the current font can be kept for the current rune. Otherwise, it calls the (probably more expensive) `ResolveFace`.)

I think this isn't a good solution either. You might keep using an exotic font after one exotic glyph. But it might be good to try staying on the current font only for spaces.

MaxGyver83:
I have updated this PR in a way that the current font is kept if the rune is a space and the font contains a space. And I have updated some tests. ~~But `TestSplit` still fails for `"الحب سماء brown привет fox تمط jumps привет over غير الأحلام"`. I don't understand why the `fox` run contains only 3 glyphs and not the space after it.~~ Now all tests pass. The first rune of " تمط " is taken from latinFont instead of arabicFont. Same for " غير الأحلام".

shaping/input.go
```diff
@@ -493,10 +494,21 @@ func ignoreFaceChange(r rune) bool {
...
 
+// isSpace returns `true` if the given rune is a space character.
+//
+// We don't want space characters to affect font selection; in general,
+// it's always wrong to select a font just to render a space.
+// We can check if a font has the ASCII space, and for other space
+// characters if they don't, HarfBuzz will compatibility-decompose them
+// to ASCII space...
+func isSpace(r rune) bool {
+	g := ucd.LookupGeneralCategory(r)
```
benoitkugler: Could you avoid calling ucd.LookupGeneralCategory twice by returning two boolean from ignoreFaceChange : `(ignore, isSpace bool)` ?

shaping/input.go
```diff
@@ -428,9 +428,16 @@ func splitByFace(input Input, availableFaces Fontmap, buffer []Input, isLast boo
...
 		// the final rune. Otherwise strings like all-whitespace are never assigned a face.
-		if ignoreFaceChange(r) && (currentInput.Face != nil || !isLast || i < input.RunEnd-1) {
-			// add the rune to the current input
-			continue
+		if currentInput.Face != nil || !isLast || i < input.RunEnd-1 {
+			if ignoreFaceChange(r) {
+				// add the rune to the current input
+				continue
+			}
+			if currentInput.Face != nil && isSpace(r) {
```
benoitkugler: Could add a comment naming the font which does not support space ?

shaping/input_test.go
```diff
@@ -139,9 +139,14 @@ func TestSplitByFontGlyphs(t *testing.T) {
 				},
 				{
 					Text:     []rune("aaa AAA "),
-					RunStart: 4, RunEnd: 8,
+					RunStart: 4, RunEnd: 7,
 					Face: upperFont,
 				},
+				{
```
benoitkugler, line 147: This test result looks wrong to me : we should not change font just for the last space.

That is because upperFont does not support spaces : I think we should change it and accept spaces.

shaping/input_test.go
```diff
@@ -222,7 +227,12 @@ func TestSplitByFontGlyphs(t *testing.T) {
 			[]Input{
 				{
 					Text:     []rune(" غير الأحلام"),
-					RunStart: 0, RunEnd: len([]rune(" غير الأحلام")),
+					RunStart: 0, RunEnd: 1,
```
benoitkugler: Same issue here : we should not change font just for the starting space

benoitkugler (changes requested):
I think I like the idea, but some tests look wrong to me.

shaping/input.go
MaxGyver83: Sure! (I even had it like this before.)

shaping/input.go
MaxGyver83: Yes, I'll add it.

shaping/input_test.go
MaxGyver83: OK, I'll change it.

shaping/input_test.go
MaxGyver83: So how to solve this? Maybe peek at the second rune to decide which font might fit? But there might be multiple spaces in a row, so this should be a loop going forward until a non-whitespace was found!?

benoitkugler: Hum... fair point.

I'm starting to think we should consider an other approach : perform the segmentation like previously, but add a second pass to check if the spaces are supported. If not, choose another font. This case will happen very rarely, so the performance penalty won't matter much.

shaping/input.go
MaxGyver83: Done in 5f3119f.

shaping/input.go
MaxGyver83: Done in 5f3119f.

shaping/input_test.go
MaxGyver83: After thinking again: Isn't it good to show what happens when a font doesn't contain a space glyph? (I think before the test was even wrong: If it actually had been rendered, it would have looked like "aaa AAA�".)

shaping/input_test.go
MaxGyver83: OK, I'll add a commit for the peek approach here. And I'll create a new PR for the two-pass approach.

MaxGyver83: Peek approach added in commit 876cd25.

MaxGyver83: > Hum... fair point.
> 
> I'm starting to think we should consider an other approach : perform the segmentation like previously, but add a second pass to check if the spaces are supported. If not, choose another font. This case will happen very rarely, so the performance penalty won't matter much.

I have started to implement this approach. I think it doesn't work as expected. If it turns out that the run contains a space which is not available in the selected font(face), you can't simply pick another font for the whole run. In my real-life bug regarding "✘ ", there is no font that has both "✘" and space (on my Android device). That's why you need different splits in such a case. If we run both the first pass and the check every time, I'm wondering what is the expensive part that is saved in the good cases.

benoitkugler: Yes, but you can create a run just for the space.

The second pass may be skipped for fonts containing a space, which is the  vast majority.

MaxGyver83: OK, I believe I need a new function for this and call it in both places before `buffer = append(buffer, currentInput)` to further split `currentInput` when necessary.

benoitkugler:
I've opened #255 to discuss the approach of a second pass.

benoitkugler:
See also #257, which seems simpler.

benoitkugler:
Closing for now, in favor of #257

PR #253, opened 2026-04-16
[internal/unicodedata] Remove previous lookup implementations

benoitkugler:
Their were used as temporary benchmark comparison, but are useless otherwise.


whereswaldon (approved):
Awesome to actually drop thousands of lines of code. :D Thanks for cleaning this up.

PR #254, opened 2026-04-17
The granular glyph data methods used a private type (gID)

rcoreilly:
From PR #220 -- switch to using GID instead of gID.


benoitkugler (approved):
Ah indeed, it is an oversight. Thank you !

PR #256, opened 2026-04-19
Fix data races in Face glyph and cmap caches

trendvidia:
## Summary

- `Face.GlyphExtents()` uses `extentsCache` (a plain slice) without synchronization
- `Face.NominalGlyph()` uses `cache21_19_8` (a plain array) without synchronization
- When a `Face` is shared across goroutines (common in GUI toolkits where layout and painting measure text concurrently), Go's race detector reports data races on both caches

This PR adds `sync.RWMutex` to both cache types: reads take an `RLock`, writes take a full `Lock`. The cache structs change from bare slice/array types to structs embedding the mutex and data.

## Changes

- `font/cache.go` — `extentsCache` becomes a struct with `sync.RWMutex` + `[]glyphExtents`; all methods take pointer receivers with proper locking
- `font/cmap_cache.go` — `cache21_19_8` becomes a struct with `sync.RWMutex` + `[256]uint32`; `get` takes pointer receiver, reads under `RLock`, writes under `Lock`
- `font/font.go` — `NewFace` uses `newExtentsCache(n)` constructor

## Test plan

- [x] `go test -race ./font/...` passes with zero race warnings
- [x] Verified in Fyne GUI toolkit test suite (`go test -race -tags ci fyne.io/fyne/v2/...`) — previously had consistent race failures in `TestCheck_Binding` and other concurrent widget tests


benoitkugler:
Hi,

Thank you for your PR. However, we are not going to add synchronisation to the font.Face object. It is is explicitly not safe for concurrent use, contrary to the font.Font struct. Adding synchronisation would reduce performance for the non concurrent use cases.

GUI toolkits should be responsible for synchronisation: either by not sharing Face (just the Font), or by wrapping it in their own structure.

benoitkugler:
See also #197

andydotxyz:
Agreed, in particular the comment in PR should not be true:

> Common in GUI toolkits where layout and painting measure text concurrently

benoitkugler:
Closing for the same reasons as in #197

PR #255, opened 2026-04-18
add a pass to handle fonts without spaces

benoitkugler:
Alternative to #248 (this is a POC and is missing tests).

PR #257, opened 2026-04-19
[shaping] do not use .notdef for non supported spaces

benoitkugler:
Almost all fonts in the wild contains a glyph for the ASCII white space, so that out font selection never selects a font just to render a space. This is in particular important to preserve correct line height (changing font could increase the line height, which is not desirable).

However, the very surprising NotoSansSymbols-Regular-Subsetted.ttf found on Android devices does *not* include a space glyph. This is very annoying and actually not so easy to fix (see previous tries at #248 and #255).

I think this PR is a better solution : it doesn't change the face selection, but replace `.notdef` glyphs by a sentinel `font.EmptyGlyph` value. Consumers of go-text are supposed to take this glyph in account for width/advance computation, but not to draw anything (that is actually the behavior of a standard space glyph).

Note that even if a toolkit does not explicitly check for the `font.EmptyGlyph`, the default behavior should be correct since the font will never have any glyph data for this glyph ID.

cc @MaxGyver83 (and @andydotxyz @whereswaldon )

### DIFF
--- a/font/renderer_test.go
+++ b/font/renderer_test.go
@@ -610,7 +610,7 @@ func TestAppleBitmapGlyph(t *testing.T) {
 
 func TestMixedGlyphs(t *testing.T) {
 	for _, filename := range tu.Filenames(t, "common") {
-		if strings.HasPrefix(filename, "common/SourceSans") {
+		if strings.HasPrefix(filename, "common/SourceSans") || strings.HasSuffix(filename, "Subsetted.ttf") {
 			continue
 		}
 		font := loadFont(t, filename)
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/go-text/typesetting
 go 1.19
 
 require (
-	github.com/go-text/typesetting-utils v0.0.0-20260327125527-fbf04b32d9ad
+	github.com/go-text/typesetting-utils v0.0.0-20260419141703-4ffe8874dabc
 	golang.org/x/image v0.23.0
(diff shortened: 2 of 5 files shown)


MaxGyver83:
I think this is a good solution! It works fine on my Android device.

One little detail: "✘ X X" is rendered like this:

X X X

The space between "✘" and the first "X" is much wider than the second space. But this shouldn't block this PR.

benoitkugler:
> One little detail: "✘ X X" is rendered like this:
> X X X
> 
> The space between "✘" and the first "X" is much wider than the second space. But this shouldn't block this PR.

Interesting ! This is probably that the first space is rendered with the first font. I'm not sure this is actually wrong in general : consider ✘ ✘ X X : which font should we use for the second space ? The first or the second ? Not clear for me there is a good, general rule.

MaxGyver83:
> Interesting ! This is probably that the first space is rendered with the first font. I'm not sure this is actually wrong in general : consider ✘ ✘ X X : which font should we use for the second space ? The first or the second ? Not clear for me there is a good, general rule.

I guess using the font of the letter before the space is most feasible.
But apart from that: In your GitHub comment (rendered in Firefox), the space between "✘" and "X" seems to be a bit wider than the space between "X" and "X" but the difference is smaller than in my Fyne screenshot.

whereswaldon (approved):
A brilliant solution to a nasty problem. Thank you @benoitkugler!

PR #258, opened 2026-04-19
[font] Tweak Cmap cache

### LINKED ISSUE #252: NominalGlyph lookups for glyphs not in the font are not cached.
For the case where we are rendering text that the font does not have support for (Say, Chinese characters in a latin-only font), having missing glyphs not cached causes performance issues. 

See https://github.com/fyne-io/fyne/pull/6263#issue-4235319448, where I attempted to fix the issue downstream in fyne. 
 The problem code is here:
```go
  // NominalGlyph returns the glyph used to represent the given rune,
  // or false if not found.
  // Note that it only looks into the cmap, without taking account substitutions
  // nor variation selectors.
  func (f *Face) NominalGlyph(ch rune) (GID, bool) {
      if g, ok := f.cmapCache.get(uint32(ch)); ok {
          return GID(g), ok
      }
      g, ok := f.Cmap.Lookup(ch)
      if ok {
          f.cmapCache.set(uint32(ch), uint32(g))
      }
      return g, ok
  }
```

benoitkugler:
Thanks to @redawl investigation, I realized our current implementation is not optimal because it passes the Cmap cache array by value, forcing a copy, while passing a pointer is more performant (see pointer.txt column).

The new benchmark results are now also better when adding a cache for not supported runes (see not-found-cache.txt column).

There is a small cost to pay when the "not supported" cache is not hit, but it is small enough.

Closes #252

|.|old.txt | pointer.txt|not-found-cache.txt|
|---|---|---|---|
||sec/op|sec/op   vs base|sec/op vs base|
| Cmap/latin_text-4    | 1004.5n ± 8% |    122.5n ± 6%  -87.81% (p=0.000 n=10) |  145.1n ± 7%  -85.55% (p=0.000 n=10)  |
| Cmap/chinese_text-4  | 769.50n ± 4% |  362.50n ± 3%  -52.89% (p=0.000 n=10)  | 93.69n ± 8%  -87.83% (p=0.000 n=10) |
| Cmap/mixed_text-4    | 1773.5n ± 3% |   488.4n ± 3%  -72.46% (p=0.000 n=10)  | 244.5n ± 3%  -86.22% (p=0.000 n=10) |
| geomean              |  1.111µ      |   278.8n       -74.90%                 | 149.2n       -86.57% |


font/cmap_cache.go
```diff
@@ -68,3 +68,34 @@ func (c *cache21_19_8) setUnchecked(key uint32, value uint32) {
 	v := (uint32(key>>8) << 19) | value
 	c[k] = v
 }
+
+// cache21_0_13 is a cache for integer keys,
+// with 0 <= key < 2097152
+type cache21_0_13 [1 << 13]uint8
```
whereswaldon, line 74: Can we document that this cache performs best on sparse keys? It generates collisions for keys less than 256 values apart. That could be useful for implementing an LRU policy when using the cache (as I think we do), but it's not immediately obvious when looking at the implementation (I had to sit here for a while and work out how this cache could manage such a large keyspace with such small storage).

whereswaldon (approved):
Thanks for tackling this! I have one non-blocking request for implementation docs, but the rest looks great.

font/cmap_cache.go
benoitkugler: Thank you, I have clarified the documentation and mentioned sparse keys.

PR #264, opened 2026-05-09
Fix spacing after color emoji

ptxmac:
When using an emoji font, space _should_ affect font selection. (e.g. Noto Emoji Color does not contain space characters).
This makes the minimal change to fix that problem, but it's likely not the best/most correct way to solve it.

Sample: https://github.com/ptxmac/gio-example/tree/test-emoji-space/headlessemoji

### DIFF
--- a/shaping/input.go
+++ b/shaping/input.go
@@ -468,12 +468,6 @@ func splitByFace(input Input, availableFaces Fontmap, buffer []Input, isLast boo
 // ignoreFaceChange returns `true` is the given rune should not trigger
 // a change of font.
 //
-// We don't want space characters to affect font selection; in general,
-// it's always wrong to select a font just to render a space.
-// We assume that all fonts have the ASCII space, and for other space
-// characters if they don't, HarfBuzz will compatibility-decompose them
-// to ASCII space...
-//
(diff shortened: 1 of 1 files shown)


benoitkugler:
Hi, 

We recently fixed a similar issue, with PR #257.
Could you instead try this and see if it solves your problem?

Otherwise, I'm afraid just ignoring spaces will break existing tests.

ptxmac:
I saw that PR, and actually thought I tested after it was merged, but must have chosen the wrong commit 🤦 

That Kinda solved the problem, though the space is now a lot bigger... But I guess that could actually be considered correct.

Visual samples using this input:
```go
const emojiSample = "🥳🧁\n🥳 🧁"
```

Old version I was trying to fix:
before

With the "fix" in this pr:
after

And how it looks on 684cf8ff69fd:
new

benoitkugler:
Thank you for trying.

I guess we could argue in both ways. 
The space (in 3) looks indeed bigger, but it is also coherent with the width of the other glyphs, looking like a monospace font.

ptxmac:
I think for my use-case, this is Good-Enough™

PR #266, opened 2026-07-09
[shaping] fix crash when wrapping part of string including forced break

benoitkugler:
We recently changed our Bidi implementation, introducing a subtle difference when handling strings containing line breaks (`\n`). 

The x/text implementation, despite claiming 

> If s contains a paragraph separator it will only process the first paragraph and report the number of bytes
> consumed from s including this separator.

actually segments `line1\nline2` as a single run containing the whole text.

The new Bidi implementation correctly recognizes two paragraphs, and according to the Unicode spec, properly truncates the text input.

This change uncovers a latent bug (see the included test). I'm not 100% sure this PR is the proper fix and would love to have confirmation from @whereswaldon

### DIFF
--- a/shaping/wrapping_test.go
+++ b/shaping/wrapping_test.go
@@ -3623,3 +3623,16 @@ func TestWrapping_oneLine_overflow_bug(t *testing.T) {
 	_, done := l.WrapNextLine(maxWidth)
 	tu.Assert(t, done)
 }
+
+func TestWrapForcedBreak(t *testing.T) {
+	f := loadOpentypeFont(t, "../font/testdata/Roboto-Regular.ttf")
+
+	s := []rune("line1\nline2")
+	// bidi segmentation will split the input at the first line
+	input := Input{Text: s, RunEnd: 5, Direction: di.DirectionLTR, Size: fixed.I(10), Face: f}
+	run := (&HarfbuzzShaper{}).Shape(input)
+
+	var l LineWrapper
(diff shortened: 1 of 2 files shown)


andydotxyz:
Can you paste the panic/crash you were seeing too - I'd love to verify if it overlaps with something I was seeing

shaping/wrapping_test.go
```diff
@@ -3623,3 +3623,16 @@ func TestWrapping_oneLine_overflow_bug(t *testing.T) {
 	_, done := l.WrapNextLine(maxWidth)
 	tu.Assert(t, done)
 }
+
+func TestWrapForcedBreak(t *testing.T) {
+	f := loadOpentypeFont(t, "../font/testdata/Roboto-Regular.ttf")
+
+	s := []rune("line1\nline2")
```
whereswaldon, line 3630: It's certainly better not to crash in this case, but isn't this API misuse? The whole shaping infrastructure expects you to perform segmentation before you shape. Segmentation should convert this input string into two runs.

Or is the problem that we previously would fail to convert this into two runs, and would shape/wrap it as a single run regardless?

I don't think I understand the full context of the problem. 😅 Sorry! I've reread the PR description a few times, but I think I'm still missing something.

benoitkugler: Actually, I think I've mixed up line break and paragraph break. That's probably why the context is not very clear here. 
We will probably need to tweak our bidi implementation, but I need to double check that. 
Sorry for the noise...

benoitkugler:
Closed in favor of #275

PR #267, opened 2026-07-13
font/cff: bound INDEX count/offSize to prevent huge allocation

mike-ward:
`parseIndexContent` calls `make([][]byte, count)` with `count` and `offSize` read directly from the font. The CFF2 INDEX header parser (`indexStart.mustParse`) doesn't validate `offSize` the way the CFF1 path (`parseIndexHeader`) does. So a malformed font can present:

- **`offSize == 0`** → `offsetArraySize` becomes 0, defeating the `len(src) < offsetArraySize` length check, or
- **a 32-bit `count` near `0xFFFFFFFF`** → `count+1` wraps to 0 in `uint32`, same effect.

Either drives a multi-gigabyte allocation before the loop errors out. A concrete real-world trigger: macOS `SFIndia.ttc` makes `font.NewFont` allocate ~58 GB, OOM-killing the process (found while shipping a pure-Go terminal that renders arbitrary Unicode).

Fix: reject `offSize` outside 1–4 and bound `count` by the bytes available for the offset array before allocating. Adds `TestParseIndexContentBounds`. Existing `font/cff` tests pass.

### DIFF
--- a/font/cff/parser.go
+++ b/font/cff/parser.go
@@ -283,22 +283,21 @@ func parseIndexContent(src []byte, header indexStart) ([][]byte, int, error) {
 	oSize := int(header.offSize)
 	// offSize must be 1..4 (CFF spec, 5176 §5). The CFF2 INDEX header parser
 	// (indexStart.mustParse) does not validate it the way the CFF1 path does,
-	// so a malformed font can present offSize 0 — which zeroes offsetArraySize
-	// below and defeats the length check — or a huge 32-bit count. Either lets
-	// make([][]byte, count) allocate gigabytes before the loop errors out.
-	// Reject a bad offSize and bound count by the bytes actually available for
-	// the offset array (each of the count+1 offsets is oSize bytes). int()
-	// widens count to avoid the uint32 wrap in count+1 (0xFFFFFFFF+1 == 0).
+	// so a malformed font can present offSize 0, which would zero the offset
+	// array size below and let the length check pass while
+	// make([][]byte, count) still allocates gigabytes.
 	if oSize < 1 || oSize > 4 {
 		return nil, 0, fmt.Errorf("reading INDEX: invalid offSize %d", oSize)
 	}
-	if int(header.count) > len(src)/oSize {
-		return nil, 0, fmt.Errorf("reading INDEX: count %d exceeds available data (%d bytes)", header.count, len(src))
+	// The offset array holds count+1 offsets of oSize bytes each. Compute in
+	// uint64 so that a 32-bit count near 0xFFFFFFFF cannot wrap, on 64- or
+	// 32-bit platforms alike. Passing this check implies the size fits in an
+	// int (it is <= len(src)) and bounds count, so the make below is safe.
+	size := (uint64(header.count) + 1) * uint64(oSize)
+	if uint64(len(src)) < size {
+		return nil, 0, fmt.Errorf("reading INDEX offsets: EOF: expected length: %d, got %d", size, len(src))
 	}
-	offsetArraySize := (int(header.count) + 1) * oSize
-	if L := len(src); L < offsetArraySize {
-		return nil, 0, fmt.Errorf("reading INDEX offsets: EOF: expected length: %d, got %d", offsetArraySize, L)
-	}
+	offsetArraySize := int(size)
 	out := make([][]byte, header.count)
 	data := src[offsetArraySize:]
 
--- a/font/cff/parser.go
+++ b/font/cff/parser.go
@@ -281,7 +281,21 @@ func parseIndexContent(src []byte, header indexStart) ([][]byte, int, error) {
 		return nil, 0, nil
 	}
 	oSize := int(header.offSize)
-	offsetArraySize := int(header.count+1) * oSize
+	// offSize must be 1..4 (CFF spec, 5176 §5). The CFF2 INDEX header parser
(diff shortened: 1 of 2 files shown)


font/cff/parser.go
```diff
@@ -281,7 +281,21 @@ func parseIndexContent(src []byte, header indexStart) ([][]byte, int, error) {
 		return nil, 0, nil
 	}
 	oSize := int(header.offSize)
-	offsetArraySize := int(header.count+1) * oSize
+	// offSize must be 1..4 (CFF spec, 5176 §5). The CFF2 INDEX header parser
```
whereswaldon, line 284: Did you evaluate whether we could instead validate this in the CFF2 INDEX header parser? Or is there a limitation there that prevents validating at parse-time?

whereswaldon (commented):
Fixing this potential huge allocation is awesome, thank you very much for tracking this problem down! I do wonder whether we should fix this in the parser though. The font parsers are @benoitkugler's area of expertise, not mine, so I'm happy to defer to his judgement about which layer should handle this issue.

font/cff/parser.go
mike-ward: I have not considered the CFF2 INDEX header parser. Good question. I'll check.

mike-ward: The CFF2 header parser is the generated `indexStart.mustParse` (`cff_gen.go`, `DO NOT EDIT`) — no error return, so it can't reject anything without changing the codegen. The `offSize` check could go in the hand-written `parseIndex2`, but the `count` bound has to stay in `parseIndexContent`: only CFF2 puts a full `uint32` in `count` (CFF1 reads a `uint16`, so `count+1` can't wrap), and by the time both paths reach `parseIndexContent` you can't tell them apart — it's also where `count+1` and the `make` actually run. One check at that shared allocation site covers CFF1, CFF2, and any future caller. Can also add the `offSize` check to `parseIndex2` for parse-time symmetry if you'd prefer.

font/cff/parser.go
```diff
@@ -281,7 +281,21 @@ func parseIndexContent(src []byte, header indexStart) ([][]byte, int, error) {
...
+	// so a malformed font can present offSize 0 — which zeroes offsetArraySize
+	// below and defeats the length check — or a huge 32-bit count. Either lets
+	// make([][]byte, count) allocate gigabytes before the loop errors out.
+	// Reject a bad offSize and bound count by the bytes actually available for
+	// the offset array (each of the count+1 offsets is oSize bytes). int()
+	// widens count to avoid the uint32 wrap in count+1 (0xFFFFFFFF+1 == 0).
+	if oSize < 1 || oSize > 4 {
+		return nil, 0, fmt.Errorf("reading INDEX: invalid offSize %d", oSize)
+	}
+	if int(header.count) > len(src)/oSize {
```
benoitkugler: Could you explain why this check (l 295) is required, given the following lines ?

benoitkugler (changes requested):
Thank you very much for the detailed report and the fix.
I have a remaining question, but the overall design looks great.

font/cff/parser.go
mike-ward: It's an overflow guard for 32-bit platforms.

mike-ward:
Knudge 😄

font/cff/parser.go
benoitkugler: Thank you.
Do you think we can drop the next check then ? It seems to be logically equivalent.

mike-ward: Not quite — they are not equivalent, and dropping the length check turns an error into a panic.

The count check floors, so it admits `count == len(src)/oSize` while `(count+1)*oSize` still exceeds `len(src)`. Concretely with `len(src)=3, oSize=1, count=3`: `3 > 3` is false so the count check passes, `offsetArraySize` is 4, and without the length check `src[4:]` panics on a 3-byte slice.

Your instinct that one check should do the job is right, though, and the pair I wrote was also still soft on 32-bit — the exact case line 295 was meant to cover. There `int` is 32 bits, so `int(0xFFFFFFFF)` is `-1`: the count check passes, `offsetArraySize` goes negative, `L < offsetArraySize` passes, and `make([][]byte, header.count)` panics with "len out of range".

So I have replaced both with a single check in uint64:

```go
size := (uint64(header.count) + 1) * uint64(oSize)
if uint64(len(src)) < size {
    return nil, 0, fmt.Errorf("reading INDEX offsets: EOF: expected length: %d, got %d", size, len(src))
}
offsetArraySize := int(size)
```

Exact rather than floored, cannot wrap on either word size, and passing it implies `size <= len(src)` so the conversion to int is safe and `count` is bounded for the `make`. One error instead of two.

The `offSize` 1..4 check has to stay separate: `offSize == 0` makes `size` zero, which satisfies any length check while `make([][]byte, count)` still allocates unbounded — that is the original `SFIndia.ttc` path.

Pushed as 4b2141b, with tests added for the floor-slack case and for `offSize > 4`.

benoitkugler (approved):
This looks great, thank you !

andydotxyz (approved):
Thanks so much :)

whereswaldon:
This is great work. Thanks so much for the detailed explanations that accompany the fix!

felix-dumit:
Could we release a new version following this fix? Also ran into this problem and would prefer not to target main directly

PR #269, opened 2026-08-18
font: avoid panic on malformed composite glyphs

ChrisJr404:
### Problem

`Face.GlyphData` / `Face.GlyphDataOutline` panic with `slice bounds out of range [:-4]` on certain malformed TrueType fonts.

`getPointsForGlyph` is documented to always return the four phantom points, so `glyphDataFromGlyf` slices them off unconditionally:

```go
points := f.getPointsForGlyph(glyph)
segments := buildSegments(points[:len(points)-phantomCount])
```

That invariant does not actually hold for composite glyphs. In `getPointsForGlyphRec`, when a component resolves to fewer than `phantomCount` points (e.g. it references an out-of-range glyph index, or nesting exceeds `maxCompositeNesting`), the function returns early *before* appending the phantom points:

```go
LC := len(compPoints)
if LC < phantomCount { // in case of max depth reached
    delete(currentGlyphs, item.GlyphIndex)
    return
}
```

So a composite glyph whose (first) component is unresolvable makes `getPointsForGlyph` return 0–3 points, and `points[:len(points)-phantomCount]` slices with a negative bound and panics. Any caller passing untrusted font data to `font.ParseTTF` and then reading glyph outlines can be crashed by a crafted font.

`getGlyfPoints` (reached from `GlyphExtents` on variable fonts) has the same unchecked assumption at `allPoints[len(allPoints)-phantomCount:]`.

### Fix

Guard both consumers: if `getPointsForGlyph` returns fewer than `phantomCount` points, treat the glyph as empty instead of slicing out of range. This matches the existing behaviour for glyphs with no outline (a blank glyph), and leaves well-formed fonts completely unaffected.

### Test

Added `TestGlyphDataMalformedComposite`, which builds a composite glyph referencing an out-of-range component and checks that the glyph-data accessors return an empty outline rather than panicking. It panics without the fix and passes with it. Found by fuzzing `ParseTTF` + glyph-outline extraction; the original crashing font and a hand-built minimal reproducer both go through the same code path.


whereswaldon:
@benoitkugler Is it better to guard the consumers, or to alter `getPointsForGlyph` to always return four points, even if they are zeroes? I don't love making all callers perform a length-checking ceremony, but maybe that is more desirable that making up phantom points? I really don't know.

benoitkugler:
@ChrisJr404 Thank you for the bug report !

> @benoitkugler Is it better to guard the consumers, or to alter `getPointsForGlyph` to always return four points, even if they are zeroes? I don't love making all callers perform a length-checking ceremony, but maybe that is more desirable that making up phantom points? I really don't know.

I indeed think it would be a bit cleaner to actually enforce the phantom points creation. It will only penalize cases which happen barely in practice, and keep the code simpler.

ChrisJr404:
Thanks, that makes sense to me too. I've reworked it so the guarantee lives in `getPointsForGlyphRec` itself: the early-return path (missing glyph id, or nesting past `maxCompositeNesting`) now appends the phantom points before returning, so `getPointsForGlyph` always yields at least `phantomCount` points and a malformed composite just produces an empty-but-valid outline. With the invariant enforced by the producer I dropped the two caller-side length checks and the redundant internal one, which keeps the code simpler as you noted. Regression test and the font/shaping/harfbuzz suites still pass.

benoitkugler (approved):
This looks great to me, thank you for this nice fix !

andydotxyz (approved):

andydotxyz:
I feel like the staticcheck issues there are some weird compile error I have seen before...
Should we just ignore?

benoitkugler:
> I feel like the staticcheck issues there are some weird compile error I have seen before... Should we just ignore?

Yes, it seems to be a go internal version mismatch. Not sure how to fix it though, perhaps we need to somehow update our CI ?

ChrisJr404:
That one is a toolchain mismatch, not anything in this change. The static_analysis job installs staticcheck v0.6.0 but the runner's `stable` Go resolved to 1.27, and the whole log is just `internal error in importing "internal/byteorder" (cannot decode ..., export data version 4 is greater than maximum supported version 2)` for stdlib packages like math/bits and internal/cpu. staticcheck v0.6.0 cannot read Go 1.27's export data format, so it bails out before it ever looks at real code. None of the errors point at font/glyphs.go or the test, and it fails the same way on unrelated branches, so the code here is clean. Looks like #272 (bump staticcheck) already covers the fix, so this is safe to ignore and should go green once CI is on a newer staticcheck.

benoitkugler:
@ChrisJr404 could you include the latest changes from main ? That should solve the CI issue.

ChrisJr404:
Merged the latest main in, which brings in the staticcheck bump from #272. That should clear the static_analysis job. Thanks!

benoitkugler:
Thank you, great job !

PR #270, opened 2026-08-20
font/cff: validate CFF2 INDEX header before sizing the object slice

felix-dumit:
## Problem

`parseIndexContent` uses `indexStart.count` to size its output slice before the
count has been validated against the input:

```go
oSize := int(header.offSize)
offsetArraySize := int(header.count+1) * oSize
if L := len(src); L < offsetArraySize {
    return nil, 0, fmt.Errorf("reading INDEX offsets: EOF: ...")
}
out := make([][]byte, header.count)
```

There are two ways to make that EOF check pass while `count` stays arbitrary:

1. **`offSize == 0`.** Then `offsetArraySize` is 0, so `len(src) < 0` is never
   true and the check is a no-op. `count` alone drives the allocation.
2. **`count == 0xFFFFFFFF`.** `header.count+1` is computed in `uint32`, so it
   wraps to 0 and zeroes `offsetArraySize` even when `offSize` is valid.

Either way `make([][]byte, count)` runs on an unvalidated `uint32`: up to
`4294967295 * 24` bytes of slice headers, roughly 103 GB. The loop that follows
then panics in `bigEndian`, which is reached with a zero-length slice
(`src[(i+1)*0 : (i+2)*0]`). A caller that recovers from the panic has still paid
for the allocation.

## Why only CFF2 is affected

`parseIndexContent` has two callers, and only one of them validates first:

- `parser.go:328` (CFF1) goes through `parseIndexHeader`, which rejects an
  `offSize` outside 1-4 and reads `count` as a `uint16`. Safe.
- `cff2.go:153` (CFF2) fills `indexStart` via `indexStart.mustParse`
  (`cff_gen.go:20-24`). That is generated code; it performs no validation and
  reads `count` as a `uint32`.

So the guarantee `parseIndexContent` relies on holds for one caller and not the
other. This puts the check in the shared function, immediately before the
allocation, so it covers both.

## Reproduction

Self-contained: synthetic bytes, no font files, no OS-specific paths. Save as
`font/cff/repro_test.go` on `v0.3.4` and run
`go test ./font/cff/ -run TestCFF2IndexOOM -v`.

```go
package cff

import (
	"encoding/binary"
	"runtime"
	"testing"
)

func TestCFF2IndexOOM(t *testing.T) {
	// A CFF2 INDEX header is 5 bytes: uint32 count, then uint8 offSize.
	// offSize 0 is invalid per 5176.CFF.pdf section 5, but indexStart.mustParse
	// does not check it, so parseIndexContent computes offsetArraySize == 0,
	// its EOF check passes, and count alone sizes the allocation.
	//
	// 10M keeps this demonstration measurable. A malformed font can supply
	// 0xFFFFFFFF, which asks for ~103 GB of slice headers.
	const count = 10_000_000

	src := make([]byte, 5, 13)
	binary.BigEndian.PutUint32(src, count)
	src[4] = 0                                // offSize
	src = append(src, 1, 2, 3, 4, 5, 6, 7, 8) // 8 bytes of body

	var before, after runtime.MemStats
	runtime.GC()
	runtime.ReadMemStats(&before)

	func() {
		defer func() {
			runtime.ReadMemStats(&after)
			if r := recover(); r != nil {
				t.Logf("panicked with: %v", r)
			}
		}()
		_, _ = parseIndex2(src, 0)
	}()

	t.Logf("input: count=%d offSize=0 len(src)=%d bytes", count, len(src))
	t.Logf("allocated: %.0f MB", float64(after.TotalAlloc-before.TotalAlloc)/(1<<20))
}
```

On `v0.3.4`, 13 bytes of input allocate 229 MB and then panic:

```
repro_test.go:32: panicked with: unreachable
repro_test.go:38: input: count=10000000 offSize=0 len(src)=13 bytes
repro_test.go:39: allocated: 229 MB
```

With this PR:

```
repro_test.go:38: input: count=10000000 offSize=0 len(src)=13 bytes
repro_test.go:39: allocated: 0 MB
```

Raising `count` to `0xFFFFFFFF` scales this to tens of GB of RSS; 10M is used
above only to keep the test cheap enough to run anywhere.

### How this shows up in practice

This is reachable from real files, not just synthetic ones. Parsing
`/System/Library/Fonts/SFIndia.ttc` (present by default on current macOS) with
`font.NewFont` drove a Go process to **14.9 GB RSS**. Each of the 9 faces in
that collection panics in `bigEndian`, having first requested between 55 GB and
92 GB. Any program that enumerates system fonts on macOS can hit this.

Because Go uses `MADV_FREE` on Darwin, the RSS is also slow to come back:
after `debug.FreeOSMemory()`, `HeapReleased == HeapSys`, yet RSS stayed at
13.9 GB. A long-lived process keeps a multi-GB floor for the rest of its life.

With this PR the same file parses cleanly: all 9 faces load, 4.4 MB allocated,
105 MB peak RSS. The bad `count`/`offSize` pair came from reading a bogus
offset, so rejecting it lets parsing continue normally rather than aborting.

## The fix

Reject an out-of-range `offSize`, and bound `count` by `len(src)` before
computing the offset array size. A well formed INDEX stores `count+1` offsets of
`offSize` bytes each, so `count < len(src)` always holds; checking that first
also keeps the multiplication from overflowing `int` on 32-bit platforms.

## Testing

- New `font/cff/index_oom_test.go`: five malformed headers are rejected, and a
  well formed INDEX still parses to the expected objects.
- `go test ./font/...` passes, including the existing real-font corpus, so the
  added validation does not reject anything that previously parsed.


felix-dumit:
Closing: this is already fixed on main by 9e0086e and 4b2141b, which I missed because I branched from the v0.3.4 tag. Upstream's single uint64 bounds check is cleaner than what I proposed, and its tests cover the same cases plus the count == len(src)/offSize boundary.

Only note left: the fix is on main but not in any tagged release, so anyone on v0.3.4 still hits multi-GB allocations parsing some system fonts (e.g. /System/Library/Fonts/SFIndia.ttc on macOS drove a process to 14.9 GB RSS). A tagged release would help downstreams pick it up without a pseudo-version.

PR #271, opened 2026-08-24
Bump version to get fix

andydotxyz:
According to https://github.com/dominikh/go-tools/issues/1711 this should fix the check errors we see


benoitkugler:
Thanks ! 
It seems the new staticcheck version reports new diagnostics. I'll try to have a look.

andydotxyz:
Updated to resolve the new issues (all in the static generator) fix is up at https://github.com/go-text/typesetting-utils/pull/28 but has test failures... I may have to leave it here for now

andydotxyz:
I cannot tell if the test failures here relate to bad internal checks, or if the variable that was supposedly doing nothing is actually being checked internally and that staticcheck was wrong.

Stumped I'm afraid but hopefully someone who knows more can cast an eye over this.

andydotxyz:
Superseded

PR #272, opened 2026-08-24
Bump staticcheck version

benoitkugler:
Supersed #271, adding lint ignore directives on generated code.


andydotxyz (commented):
In other repos I work on committing commended out code is not OK - is that acceptable here?

The other question I have is it looks like the ignores have been added in the generated output but not the generator itself, so won't they go away on next generation call?

benoitkugler:
> In other repos I work on committing commended out code is not OK - is that acceptable here?

I am really sorry, but I did not grasp the meaning of "commended out". 

> The other question I have is it looks like the ignores have been added in the generated output but not the generator itself, so won't they go away on next generation call?

Fair point. I have opened https://github.com/go-text/typesetting-utils/pull/29 for the generator change itself.

andydotxyz:
> I am really sorry, but I did not grasp the meaning of "commended out".

A typo sorry - I meant "commented out". Normally git is sufficient to track old code if it's needed in the future.

benoitkugler:
> > I am really sorry, but I did not grasp the meaning of "commended out".
> 
> A typo sorry - I meant "commented out". Normally git is sufficient to track old code if it's needed in the future.

Ah, thank you for the clarification. I've removed the commented code.

benoitkugler:
@andydotxyz friendly ping, so that we can fix the CI and publish a new tag

andydotxyz (approved):
Sorry, I'm drowning in notifications, approved now

PR #274, opened 2026-09-03
font/cff: resolve CFF2 local subrs relative to the Private DICT

anaelorlinski:
The `Subrs` operand in a Private DICT is an offset *"relative to the beginning of the
Private DICT data"* (5176.CFF.pdf §15, CFF2 §10). `ParseCFF2` was treating it as an
offset from the start of the table, so any CFF2 font declaring local subroutines had its
Local Subrs INDEX read from an arbitrary position.

The CFF1 parser already handles this correctly, in `parsePrivateDict`:

```go
// "The local subrs offset is relative to the beginning of the Private DICT data"
if err = p.seek(offset + priv.subrsOffset); err != nil {
```

The CFF2 path omitted the base:

```go
-out.fonts[i].localSubrs, err = parseIndex2(src, int(pd.subrsOffset))
+subrsOffset := int(fd.privateDictOffset) + int(pd.subrsOffset)
+out.fonts[i].localSubrs, err = parseIndex2(src, subrsOffset)
```

### What the wrong position yields

Whatever happens to be at that address. On this repo's own `toys/CFF2-VF.otf` it is a
byte giving `offSize 140`, so the table is rejected. Of the three CFF2 fonts shipped with
macOS that declare local subroutines, two present counts of 28 and 58 million and fail the
length check, and `ZitherIndia.otf` presents `offSize 0` — which before #267 fell through
to `bigEndian(src[0:0])` and panicked with `"unreachable"`, and now errors.

For `ZitherIndia.otf` the arithmetic is easy to see: its Private DICT is at 1206785 with
size 161, and `Subrs` is 161 (the INDEX sits immediately after the DICT, as usual).

| | address | count | offSize |
|---|---|---|---|
| read as-is, `src[161]` | 161 | 2717909960 | **0** |
| relative to the Private DICT, `src[1206785+161]` | 1206946 | 17582 | 3 |

### Why it went unnoticed

Two reasons, both worth fixing separately at some point:

1. `NewFont` discards the error — `out.cff2, _ = loadCff2(...)` — so an affected font
   loads with **no CFF2 outlines** rather than failing.
2. No test reached the code. The only other CFF2 fixture,
   `common/NotoSansCJKjp-VF.otf`, has 18 font DICTs and **zero** local subroutines. And
   the existing user of `toys/CFF2-VF.otf` (`TestVar`) exercises `fvar` normalization
   only — it never asks for a glyph, so it passes while the font's outlines are absent.

### After

All four parse, and every glyph interprets:

| font | before | after |
|---|---|---|
| `toys/CFF2-VF.otf` | `invalid offSize 140` | 5 charstrings, 3 local subrs |
| `ZitherIndia.otf` | `invalid offSize 0` (panic before #267) | 5683 charstrings, 17582 local subrs |
| `ZitherMalayalam.otf` | `EOF: expected length 232001036` | 198 charstrings, 485 local subrs |
| `ZitherTamil.otf` | `EOF: expected length 28246451` | 647 charstrings, 770 local subrs |

`common/NotoSansCJKjp-VF.otf` is unaffected either way — it has no local subroutines.

### Test

`TestCFF2LocalSubrs` uses the existing `toys/CFF2-VF.otf`, so no new fixture is needed. It
asserts the font declares subroutines and that every glyph loads — the latter is what
actually exercises them, since a charstring reaching `callsubr` with a wrongly located
INDEX fails there. Without this change it fails with
`reading INDEX: invalid offSize 140`.

### Also included

A guard against a negative offset in `parseIndex2`. Its length check
(`if L := len(src); L < offset+5`) passes trivially for a negative offset, and
`src[offset:]` then panics. This is newly reachable now that the offset is the sum of two
values read from the font.

### Relation to #267

#267 added the `offSize` validation whose comment notes that *"The CFF2 INDEX header
parser does not validate it the way the CFF1 path does"*. This is the other half of that
story: the reason a perfectly valid font was presenting a bogus `offSize` in the first
place. #267 stopped the crash; this stops the misread that caused it.

### DIFF
--- a/font/cff/cff2.go
+++ b/font/cff/cff2.go
@@ -98,7 +98,13 @@ func ParseCFF2(src []byte) (*CFF2, error) {
 		out.fonts[i].defaultVSIndex = pd.vsindex
 		// if required, parse the local subroutines
 		if pd.subrsOffset != 0 {
-			out.fonts[i].localSubrs, err = parseIndex2(src, int(pd.subrsOffset))
+			// "The local subrs offset is relative to the beginning of the
+			// Private DICT data" (5176.CFF.pdf section 15, and CFF2 section
+			// 10), so it is measured from the Private DICT, not from the start
+			// of the table. The CFF1 parser does this correctly; see
+			// parsePrivateDict.
+			subrsOffset := int(fd.privateDictOffset) + int(pd.subrsOffset)
+			out.fonts[i].localSubrs, err = parseIndex2(src, subrsOffset)
 			if err != nil {
 				return nil, err
 			}
@@ -143,6 +149,9 @@ func ParseCFF2(src []byte) (*CFF2, error) {
 }
 
 func parseIndex2(src []byte, offset int) ([][]byte, error) {
+	if offset < 0 {
+		return nil, fmt.Errorf("reading INDEX: invalid offset %d", offset)
+	}
 	if L := len(src); L < offset+5 {
 		return nil, fmt.Errorf("reading INDEX: EOF: expected length: %d, got %d", offset+5, L)
 	}
--- a/font/cff/parser_test.go
+++ b/font/cff/parser_test.go
@@ -156,6 +156,47 @@ func TestIssue122(t *testing.T) {
 	tu.Assert(t, len(segments) == 12)
 }
 
+// TestCFF2LocalSubrs covers a CFF2 font whose Private DICT declares local
+// subroutines. The Subrs offset is relative to the Private DICT, not to the
+// start of the table, so resolving it from the wrong base reads an arbitrary
+// position: with this fixture that lands on a byte giving offSize 140 and the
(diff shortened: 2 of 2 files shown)


benoitkugler:
That looks really great, thank you for the fix and the detailed diagnostics. 

Is there a way you could include the ZitherXXX fonts in a test, either by extracting the CFF2 table or by including the fonts in the dev repository `go-text/typesetting-utils` ? (If the license allows it.)

anaelorlinski:
The font is part of macOS so it's not redistributable and probably not a good idea to include it. Anyway I made Claude generate a synthetic test. I can still send the font by email if you need.

benoitkugler (approved):
Thank you !

benoitkugler:
Merging, since it is a bug fix, with no API change.

PR #275, opened 2026-09-16
[bidi,shaping] better handle input containing multiple paragraphs

benoitkugler:
This supersed #266 , with a better solution : we properly split out the text input in paragraphs (delimited by bidi class B, like `\n`) and we apply the BIDI algorithm on each paragraph.

There is no change when wrapping lines : the new runs are simply created earlier.

This fixes a bug I have downstream when wrapping text like `"line1\nline2"`.


whereswaldon (approved):
In general, I'd expect applications/toolkits to segment by paragraph before ever invoking our logic. However, it does little harm to ensure that we handle this gracefully. Thank you for implementing it!

PR #276, opened 2026-09-17
[shaping] use BIDI levels to properly compute visual ordering

### LINKED ISSUE #249: [shaping] Bidi reordering for nested RTL/LTR text

benoitkugler:
This PR implements the BIDI re-ordering algorithm, handling nested BIDI contents.

Under the hood, we propagate the BIDI level computed when segmenting the input, and use it after line wrapping.

There are two related question we should answer : 

1. Is the `Direction` field of the line wrapping config still required ?
2. Should we expose the level field, so that consumer which do not use our segmentation (`Split`) may still have correct results after line wrapping ?

Fixes #249


andydotxyz:
This looks good, as for the questions:


> Is the Direction field of the line wrapping config still required ?

As far as I understand we may still want to pass in that context, rather than assuming the embedding / run is operating always from the outer-most level.

> Should we expose the level field, so that consumer which do not use our segmentation (Split) may still have correct results after line wrapping ?

I'm not sure about this as I've not used the wrapping from this codebase yet - but if we expose the same field on other structures i'd lean towards thinking it's appropriate to expose...

benoitkugler:
> I'm not sure about this as I've not used the wrapping from this codebase yet - but if we expose the same field on other structures i'd lean towards thinking it's appropriate to expose...

Thank you ! 
I've exposed the `Level` fields and added some documentation.

benoitkugler:
Regarding `config.Direction` field, it is actually still used. The thing is : this field should always match the `Input.Direction` the user provide when segmenting its input (before shaping and wrapping happens), so that the `Level`s computed by the BIDI algorithm properly match the wrapping direction.
So I guess the status quo is the best option; and I have tried to add documentation to clarify this dependency.

whereswaldon (approved):
Wow, this was so much more straightforward to implement than I had anticipated! Thank you so much for taking this on. It's such a relief that we can implement this in the line post-processing instead of needing to modify the wrapping logic with this additional constraint.

benoitkugler:
> Wow, this was so much more straightforward to implement than I had anticipated! Thank you so much for taking this on. It's such a relief that we can implement this in the line post-processing instead of needing to modify the wrapping logic with this additional constraint.

You're welcome ! The BIDI algorithm indeed is rather clear and states that bidi reordering happens *after* line breaking 😄

benoitkugler:
(Waiting for @andydotxyz final review since there is an API change.)

PR #277, opened 2026-09-21
[opentype] expose flavor when writting a font file

benoitkugler:
One typical usage of `font/opentype.WriteTTF` is for font subsetting and inclusion in PDF files. 

This issue is that some PDF viewers will not render glyphs from a 'CFF' table if the font file is tagged with the 'TrueType' flavor (aka sfntVersion).

The fix is straightforward : simply copy the snftVersion of the font you want to subset. The issue is that our 
`opentype.WriteTTF` function does not expose this parameter : it always uses 'TrueType'.

This PR provides a new function `WriteOpentype` which lets the user decide which flavor to use. I've also deprecated `WriteTTF` since the replacement is straightforward.


whereswaldon (approved):
Change and rationale make sense to me. Thank you.

